PR #12, opened 2020-04-21
Hash seed passed to GeneratePrivateKey using SHA3_384

psiemens:
This is adapted from: https://github.com/dapperlabs/flow-go-sdk/commit/6b40389055ca30d72c7bdad3cc6def7137c5f540

cc @tarakby


crypto/crypto.go
```diff
@@ -43,16 +44,22 @@ func (f SignatureAlgorithm) String() string {
...
 }
 
-// MinSeedLength returns the minimum seed length that can safely be used to generate private keys with this algorithm.
-func (f SignatureAlgorithm) MinSeedLength() int {
+// MinSeedLength is a generic minimum seed length to guarantee a minimum entropy.
+// It is used when the seed source is not necessary a CSPRG and the seed
+// should be expanded before being passed to the key generation process.
+const MinSeedLength = crypto.MinSeedLen
+
+// minSeedLength returns the minimum seed length that can safely be used to generate private keys with this algorithm.
```
tarakby, line 52: I think we might not need this function as it's only used in one place. Feel free to remove it or not

crypto/crypto.go
```diff
@@ -43,16 +44,22 @@ func (f SignatureAlgorithm) String() string {
 	return [...]string{"UNKNOWN", "BLS_BLS12381", "ECDSA_P256", "ECDSA_secp256k1"}[f]
 }
 
-// MinSeedLength returns the minimum seed length that can safely be used to generate private keys with this algorithm.
-func (f SignatureAlgorithm) MinSeedLength() int {
+// MinSeedLength is a generic minimum seed length to guarantee a minimum entropy.
+// It is used when the seed source is not necessary a CSPRG and the seed
+// should be expanded before being passed to the key generation process.
+const MinSeedLength = crypto.MinSeedLen
```
tarakby, line 50: This is just exporting an internal constant. Feel free to update it.

crypto/crypto.go
```diff
@@ -42,6 +44,24 @@ func (f SignatureAlgorithm) String() string {
...
+func (f SignatureAlgorithm) minSeedLength() int {
+	switch f {
+	case ECDSA_P256:
+		return crypto.KeyGenSeedMinLenECDSAP256
+	case ECDSA_secp256k1:
+		return crypto.KeyGenSeedMinLenECDSASecp256k1
+	}
+
+	// return a high seed length
+	return 1<<31 - 1
```
psiemens, line 62: Hmm, I like the idea to return a high number, but this might cause users to inadvertently allocate 4GB of memory 😬

crypto/crypto.go
```diff
@@ -215,7 +230,27 @@ func NewNaiveSigner(privateKey PrivateKey, hashAlgo HashAlgorithm) NaiveSigner {
...
+			len(seed),
+			MinSeedLength,
+			sigAlgo,
+		)
+	}
+
+	// expand the seed and uniformize its entropy
+	generationTag := []byte("ECDSA Key Generation")
+	customizer := []byte("")
+	hasher, err := hash.NewKMAC_128(generationTag, customizer, sigAlgo.minSeedLength())
```
psiemens, line 246: Maybe we could remove `sigAlgo.minSeedLength()` and instead put this inline:

```go
        var seedLen int
	switch f {
	case ECDSA_P256:
		seedLen = crypto.KeyGenSeedMinLenECDSAP256
	case ECDSA_secp256k1:
		seedLen = crypto.KeyGenSeedMinLenECDSASecp256k1
        default:
                 return PrivateKey{}, fmt.Errorf("crypto: invalid signature algorithm %s", sigAlgo)
	}
```

crypto/crypto.go
```diff
@@ -215,7 +230,27 @@ func NewNaiveSigner(privateKey PrivateKey, hashAlgo HashAlgorithm) NaiveSigner {
 
 // GeneratePrivateKey generates a private key with the specified signature algorithm from the given seed.
 func GeneratePrivateKey(sigAlgo SignatureAlgorithm, seed []byte) (PrivateKey, error) {
-	privKey, err := crypto.GeneratePrivateKey(crypto.SigningAlgorithm(sigAlgo), seed)
+	// check the seed has minimum entropy
+	if len(seed) < MinSeedLength {
```
psiemens, line 234: Should this check `sigAlgo.minSeedLength()`, or does it need to be the generic length?

crypto/crypto_test.go
```diff
@@ -0,0 +1,103 @@
...
+	equalSeed := makeSeed(crypto.MinSeedLength)
+	longSeed := makeSeed(crypto.MinSeedLength * 2)
+
+	t.Run("Invalid signature algorithm", func(t *testing.T) {
+		sk, err := crypto.GeneratePrivateKey(invalidAlgo, longSeed)
+		assert.Error(t, err)
+		assert.Equal(t, crypto.PrivateKey{}, sk)
+	})
+
+	for _, validAlgo := range []crypto.SignatureAlgorithm{crypto.ECDSA_P256, crypto.ECDSA_secp256k1} {
```
psiemens, line 44: 👏 Great idea!

crypto/crypto.go
tarakby: lol you're right, I have made that change at the beginning when the seed length was compared to the output of this function. (although the user allocation will be saved by this 😄  https://github.com/onflow/flow-go-sdk/pull/12/files#diff-279ba67e7dec0c2c6d76346444e8d179R53 )

crypto/crypto.go
tarakby: Yeah I agree

crypto/crypto.go
tarakby: That needs to be the generic length. 
The specific one is done to the length of hash seed here `privKey, err := crypto.GeneratePrivateKey(crypto.SigningAlgorithm(sigAlgo), hashedSeed)` and that's guaranteed because KMAC has an arbitrary output, and we precised that we need an output length `sigAlgo.minSeedLength()`

crypto/crypto.go
```diff
@@ -234,15 +238,20 @@ func GeneratePrivateKey(sigAlgo SignatureAlgorithm, seed []byte) (PrivateKey, er
 		seedLen = crypto.KeyGenSeedMinLenECDSAP256
 	case ECDSA_secp256k1:
 		seedLen = crypto.KeyGenSeedMinLenECDSASecp256k1
-        default:
-                 return PrivateKey{}, fmt.Errorf("crypto: invalid signature algorithm %s", sigAlgo)
+	default:
+		return PrivateKey{}, fmt.Errorf(
+			"crypto: key generation not supported for non-ECDSA algorithm %s",
```
tarakby, line 243: I think a more precise error message would be that the signature algorithm is just not supported by the SDK. Why: the SDK only supports ECDSA with 2 curves, so we could have other curves implemented with ECDSA but they are still not supported by the SDK.

psiemens: Great point, I'll add 👍

crypto/crypto_test.go
```diff
@@ -28,69 +28,94 @@ import (
...
 func TestGeneratePrivateKey(t *testing.T) {
+	ecdsaAlgos := []crypto.SignatureAlgorithm{
+		crypto.ECDSA_P256,
+		crypto.ECDSA_secp256k1,
+	}
+
+	nonECDSAAlgos := []crypto.SignatureAlgorithm{
+		crypto.BLS_BLS12381,
+	}
+
```
tarakby, line 39: This is minor, but same as above, the separation is more about SDK-supported/non-SDK-supported rather than ECDSA/non-ECDSA.

crypto/crypto_test.go
```diff
@@ -28,69 +28,94 @@ import (
...
 
 func TestGeneratePrivateKey(t *testing.T) {
+	ecdsaAlgos := []crypto.SignatureAlgorithm{
+		crypto.ECDSA_P256,
+		crypto.ECDSA_secp256k1,
+	}
+
+	nonECDSAAlgos := []crypto.SignatureAlgorithm{
+		crypto.BLS_BLS12381,
+	}
```
tarakby, line 38: Also minor: I think from the SDK point of view, it expects the same behavior from non-supported algorithms or non-valid algorithms, so I guess `crypto.SignatureAlgorithm(-42)` could even be included in this array. 

It's just a comment to point this out, but I let you decide how to implement this test.

crypto/crypto_test.go
```diff
@@ -28,69 +28,94 @@ import (
...
 	shortSeed := makeSeed(crypto.MinSeedLength / 2)
 	equalSeed := makeSeed(crypto.MinSeedLength)
 	longSeed := makeSeed(crypto.MinSeedLength * 2)
 
-	t.Run("Invalid signature algorithm", func(t *testing.T) {
-		sk, err := crypto.GeneratePrivateKey(invalidAlgo, longSeed)
-		assert.Error(t, err)
-		assert.Equal(t, crypto.PrivateKey{}, sk)
-	})
+	for _, sigAlgo := range ecdsaAlgos {
```
tarakby, line 47: Cleaner 👏

crypto/crypto_test.go
```diff
@@ -28,69 +28,94 @@ import (
 )
 
 func TestGeneratePrivateKey(t *testing.T) {
+	ecdsaAlgos := []crypto.SignatureAlgorithm{
```
tarakby, line 31: cleaner 👏

crypto/crypto_test.go
psiemens: I intentionally treated this as a separate case because `-42` does not represent a valid algorithm in any sense, whereas BLS is valid but _not supported_ by the SDK. I updated the test to be more clear about this distinction.

tarakby: I see your point 👌

PR #23, opened 2020-04-30
Include account key IDs in protobuf messages

psiemens:
Ref: https://github.com/dapperlabs/flow-emulator/issues/46

## Description

Account key IDs were being omitted from protobuf messages. This PR fixes the `convert` package to include them.

### DIFF
--- a/client/convert/protobuf.go
+++ b/client/convert/protobuf.go
@@ -358,6 +358,7 @@ func MessageToAccountKey(m *entities.AccountKey) (*flow.AccountKey, error) {
 	}
 
 	return &flow.AccountKey{
+		ID:             int(m.GetIndex()),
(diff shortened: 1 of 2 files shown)


m4ksio:
Do we need this? Key ID/Index is about it's position in the account, and not necessary part of the account key itself.

psiemens:
@m4ksio Yeah, we actually do need it now.

We decided to preserve the key indices of all remaining keys when a key is removed from an account, which means a key's index isn't necessarily its position in the returned list. This is mainly to prevent bugs in which somebody may reference a key that later changes indices. As a result, we also decided to rename `key index` => `key ID` to avoid confusion, since index implies positioning whereas ID implies uniqueness.

PR #35, opened 2020-05-22
Set the algo when setting the pub key

Kay-Zee:
## Description

Sets the `AccountKey` Signature field automatically when setting the public key


account.go
```diff
@@ -60,6 +60,7 @@ func (a *AccountKey) FromPrivateKey(privKey crypto.PrivateKey) *AccountKey {
 // SetPublicKey sets the public key for this account key.
```
psiemens, line 60:
```suggestion
// SetPublicKey sets the public key for this account key.
//
// This function automatically sets the the SigAlgo field based on the provided key.
```

PR #38, opened 2020-05-26
Update examples to be compatible with new emulator

psiemens:
## Description

This PR updates the emulator examples to be compatible with the upcoming emulator release.


examples/examples.go
```diff
@@ -46,7 +46,7 @@ func ServiceAccount(flowClient *client.Client) (flow.Address, *flow.AccountKey,
 	privateKey, _ := crypto.GeneratePrivateKey(crypto.ECDSA_P256, []byte(defaultServiceKeySeed))
 
 	// Service address is the first generated account address
```
turbolent, line 48: Maybe point out that this needs to be changed if run against mainnet

psiemens: Good idea! I'm going to follow up with more extensive documentation on the new address format and how to switch between chains.

PR #41, opened 2020-05-28
Update flow.AddressGenerator API

psiemens:
## Description

This PR attempts to make the `flow.AddressGenerator` a bit more declarative. 

The old API is still supported, but you can now do the following:

```go
gen := flow.NewAddressGenerator(flow.Mainnet)

// get the current address
addressA := gen.Address()

// increment state
gen.Next()

// get the next address
addressB := gen.Address()

// skip to index 42
gen.SetIndex(42)

addressC := gen.Address()
```

I also removed the error return values, with the reason being that the only error case is when `state > maxState`. Because `maxState` is so large, it feels a bit unnecessary to handle an error that will never occur under normal operation. Chances are if somebody is trying to generate past the very last address, they're already hitting some other logical limit in their application (especially users of the SDK).

This allows for chaining like this:

```go
address42 := flow.NewAddressGenerator().SetIndex(42).Address()
```

Other changes:
- Bring back `test.AddressGenerator` for consistency with other test entity generators
- Add `flow.EmptyAddress`


address.go
```diff
@@ -48,46 +55,48 @@ func newAddressGeneratorAtState(chainID ChainID, state addressState) *AddressGen
...
 // addressState represents the internal state of the address generation mechanism
 type addressState uint64
 
 // AddressLength is the size of an account address in bytes.
-// linearCodeN is the size of an account address in bits.
 const AddressLength = (linearCodeN + 7) >> 3
 
 const (
-	// ZeroAddressState is the addressing state when Flow is bootstrapped
+	// seroAddressState is the addressing state when Flow is bootstrapped
```
tarakby, line 97:
```suggestion
	// zeroAddressState 
```

address.go
```diff
@@ -221,11 +233,8 @@ func generateAddress(chain ChainID, state addressState) Address {
 // and false otherwise.
 //
 // This is an off-chain check that only tells whether the address format is
-// valid. If the function returns true, this does not mean
-// a Flow account with this address has been generated. Such a test would
-// require on on-chain check.
-// ZeroAddress fails the check. Although it has a valid format, no account
-// in Flow is assigned to ZeroAddress.
+// valid. If the function returns true, this does not mean a Flow account with
+// this address has been generated. Such a test would require on on-chain check.
```
tarakby, line 237:
```suggestion
// this address has been generated. Such a test would require an on-chain check.
```

tarakby, line 237:
```suggestion
// this address has been generated. Such a test would require an on-chain check.
```

address.go
tarakby, line 97:
```suggestion
	// zeroAddressState is the addressing state when Flow is bootstrapped
```

address.go
```diff
@@ -48,46 +55,48 @@ func newAddressGeneratorAtState(chainID ChainID, state addressState) *AddressGen
...
-	gen.state += 1
-	return nil
+
+	gen.state++
+
+	return gen
+}
+
+// SetIndex fast-forwards or rewinds the addressing state to the given index.
+func (gen *AddressGenerator) SetIndex(i uint) *AddressGenerator {
```
tarakby, line 85: I think the idea of removing the error in `Next` assumes that it's unpractical to reach the index `2^45`, but with exposing this API, a caller can jump to that index and wouldn't get any error.  

Another comment: shouldn't a function `AddressAtIndex(i) Address` be one that a developer would use rather than `SetIndex(i) * AddressGenerator `. In both cases I feel like developers will use this API and skip all the others :)

tarakby: Anyway, I don't have a strong opinion about the API, you can set the interface you think is better. 
I didn't set it this way as 3 weeks ago the design of addresses didn't work with indices at all :)

address.go
```diff
@@ -48,46 +55,48 @@ func newAddressGeneratorAtState(chainID ChainID, state addressState) *AddressGen
...
+	return generateAddress(gen.chainID, gen.state)
 }
 
-// nextState increments the addressing state.
+// Next increments the addressing state.
 //
-// The state values are incremented from 0 to 2^k-1
-func (gen *AddressGenerator) nextState() error {
+// State values are incremented from 0 to 2^k-1.
+func (gen *AddressGenerator) Next() *AddressGenerator {
```
tarakby, line 72: Just pointing out that having `Next` and `Address` together can replace a call to `NextAddress`. Maybe we can keep only one option to developers not to get them confused?

psiemens: I'm going to keep `NextAddress` in order to stay backwards compatible for the next few releases, but I agree that it should be deprecated in the future.

PR #42, opened 2020-05-30
Add timestamp to block header

philipstanislaus:
Required for https://github.com/dapperlabs/flow-go/issues/3271 and https://github.com/dapperlabs/flow-go/issues/2255


block.go
```diff
@@ -18,23 +18,26 @@
...
+	Height    uint64
+	Timestamp time.Time
 }
 
-// A BlockPayload is the full contents of a block.
+// Payload is the full contents of a block.
 //
 // A payload contains the collection guarantees and seals for a block.
-type BlockPayload struct {
+type Payload struct {
```
psiemens, line 40: I'd prefer to keep these as `BlockHeader` and `BlockPayload`, just because they exist in the same package as `Transaction`, which also has a payload, and I'd like to be explicit (and a bit more verbose) to avoid confusion about which entity they belong to.

client/convert/protobuf.go
```diff
@@ -104,54 +105,78 @@ func MessageToAccountKey(m *entities.AccountKey) (*flow.AccountKey, error) {
...
-func BlockToMessage(b flow.Block) *entities.Block {
-	return &entities.Block{
-		Id:                   b.ID.Bytes(),
-		ParentId:             b.ParentID.Bytes(),
-		Height:               b.Height,
-		CollectionGuarantees: CollectionGuaranteesToMessages(b.CollectionGuarantees),
+func BlockToMessage(b flow.Block) (*entities.Block, error) {
+	t, err := ptypes.TimestampProto(b.Header.Timestamp)
+	if err != nil {
+		return nil, err
```
psiemens, line 111:
```suggestion
		return nil, fmt.Errorf("convert: failed to convert block timestamp to message: %w", err)
```

client/convert/protobuf.go
```diff
@@ -104,54 +105,78 @@ func MessageToAccountKey(m *entities.AccountKey) (*flow.AccountKey, error) {
...
 
-func BlockHeaderToMessage(b flow.BlockHeader) *entities.BlockHeader {
-	return &entities.BlockHeader{
-		Id:       b.ID.Bytes(),
-		ParentId: b.ParentID.Bytes(),
-		Height:   b.Height,
+func BlockHeaderToMessage(b flow.Header) (*entities.BlockHeader, error) {
+	t, err := ptypes.TimestampProto(b.Timestamp)
+	if err != nil {
+		return nil, err
```
psiemens, line 154:
```suggestion
		return nil, fmt.Errorf("convert: failed to convert message to block timestamp: %w", err)
```

block.go
philipstanislaus: Okay, changed back.

block.go
```diff
@@ -18,20 +18,23 @@
...
 package flow
 
-// A Block is a set of state mutations applied to the Flow blockchain.
+import "time"
+
+// Block is a set of state mutations applied to the Flow blockchain.
 type Block struct {
-	BlockHeader
-	BlockPayload
+	BlockHeader  *BlockHeader
```
psiemens, line 25: Sorry, I missed this in my last review. Why change these from embedded structs to fields?

philipstanislaus: Reverted, thanks!

PR #43, opened 2020-06-01
Transaction arguments

psiemens:
## Description

This PR adds SDK support for transaction arguments.

- Add `Arguments` field to `flow.Transaction`
- Add `AddArgument` method to `flow.Transaction`
- Include arguments in protobuf messages, update convert functions & tests

### DIFF
--- a/test/greetings.go
+++ b/test/greetings.go
@@ -1,0 +1,140 @@
+/*
+ * Flow Go SDK
+ *
+ * Copyright 2019-2020 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package test
+
+var GreetingScript = []byte(`transaction(greeting: String) { execute { log(greeting + ", World!") } }`)
+
+const (
+	greetingMsgAF = "Hi"
+	greetingMsgSQ = "Hi"
+	greetingMsgAM = "ሃይ"
+	greetingMsgAR = "مرحبا"
+	greetingMsgEU = "Hi"
+	greetingMsgBN = "হাই"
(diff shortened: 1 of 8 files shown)


transaction.go
```diff
@@ -222,6 +232,11 @@ func (t *Transaction) payloadCanonicalForm() interface{} {
 		authorizers[i] = auth.Bytes()
 	}
 
+	arguments := make([][]byte, len(t.Arguments))
+	for i, arg := range t.Arguments {
+		arguments[i], _ = jsoncdc.Encode(arg)
```
turbolent, line 237: Is the second return value here an error? If so should maybe be returned?

Or is it assumed the arguments are all valid here? Maybe then assert/panic if they aren't?

Don't quite know what this `payloadCanonicalForm` function is used for

psiemens: Yeah, it's an error. 

> Or is it assumed the arguments are all valid here?

Unless there is a bug in our code or somebody manually constructs an invalid `cadence.Value` object, it shouldn't be possible to pass an invalid value. I struggle a bit with cases like this: do we inconvenience the happy path (ID generation, hashing, signatures) in order to catch a case that only occurs under misuse? 

For now I'd like to add a panic so that it doesn't fail silently, but I also want to give some more thought to how we handle and propagate these kinds of errors.

test/greetings.go
```diff
@@ -0,0 +1,140 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package test
+
+var GreetingScript = []byte(`transaction(greeting: String) { execute { log(greeting + ", World!") } }`)
+
+const (
```
turbolent, line 23: ❤️

transaction.go
turbolent: Right, I think the errors produced from the encoding, like composites with mismatching field values and field types, are "just" going to cause invalid transactions, which should be fine. 

I guess the error can just be bubbled up, but like you wrote, maybe should be thought through more and done in a separate PR 👍

transaction.go
```diff
@@ -222,6 +233,18 @@ func (t *Transaction) payloadCanonicalForm() interface{} {
...
 	}
 
+	var err error
+
+	arguments := make([][]byte, len(t.Arguments))
+	for i, arg := range t.Arguments {
+		arguments[i], err = jsoncdc.Encode(arg)
+		if err != nil {
+			panic(
+				fmt.Sprintf("failed to encode argument at index %d: %s", i, err.Error()),
```
turbolent, line 243: 👍

PR #47, opened 2020-06-04
client/convert: Skip decoding of block timestamp if nil

### LINKED ISSUE #46: Block and BlockHeader methods not compatible with current mainnet release
### Problem

The Flow Go SDK `v0.4.0` breaks when calling the `GetBlock...` and `GetBlockHeader` methods of the Access API currently exposed by candidate mainnet.

### Steps to Reproduce 

For example, the following error is returned when querying `GetLatestBlockHeader`:

```
GetLatestBlockHeader: client: timestamp: nil Timestamp
```
  
### Context

This happens because the current deployed Access API does not include a timestamp in the block header response.

psiemens:
Closes #46

## Description

This updates the `Block` and `BlockHeader` client methods to be compatible with versions of the Access API that do no implement the `BlockHeader.Timestamp` field.

### DIFF
--- a/test/entities.go
+++ b/test/entities.go
@@ -142,26 +142,29 @@ func (g *Blocks) New() *flow.Block {
 }
 
 type BlockHeaders struct {
-	count    int
-	ids      *Identifiers
-	lastTime time.Time
+	count     int
+	ids       *Identifiers
+	startTime time.Time
 }
 
 func BlockHeaderGenerator() *BlockHeaders {
+	startTime, _ := time.Parse(time.RFC3339, "2020-06-04T15:43:21+00:00")
+
 	return &BlockHeaders{
-		count:    1,
-		ids:      IdentifierGenerator(),
-		lastTime: time.Now().UTC(),
+		count:     1,
+		ids:       IdentifierGenerator(),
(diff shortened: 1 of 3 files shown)


test/entities.go
```diff
@@ -142,26 +142,29 @@ func (g *Blocks) New() *flow.Block {
...
-	count    int
-	ids      *Identifiers
-	lastTime time.Time
+	count     int
+	ids       *Identifiers
+	startTime time.Time
 }
 
 func BlockHeaderGenerator() *BlockHeaders {
+	startTime, _ := time.Parse(time.RFC3339, "2020-06-04T15:43:21+00:00")
```
turbolent, line 151: This is great, just curious: Is there a particular reason why the test changed? Is it to make it deterministic?

psiemens: Yep! All of the other entity generators are deterministic in that they produce the exact same sequence of entities on each run. I realized that `time.Now` breaks determinism for the `Blocks` and `BlockHeaders` generators, so I made this change to make them consistent with the rest. It's not a strict requirement but I want to preserve consistency. :)

Thanks for the reviews BTW 🙇

PR #49, opened 2020-06-05
Update emulator examples to read service key from flow.json

psiemens:
This PR introduces a small improvement to the `examples` package that loads the service account information from `flow.json` rather than a hard-coded seed.

### DIFF
--- a/examples/examples.go
+++ b/examples/examples.go
@@ -62,7 +62,15 @@ type config struct {
 
 func readConfig() config {
 	f, err := os.Open(configPath)
-	Handle(err)
+	if err != nil {
+		if os.IsNotExist(err) {
+			fmt.Println("Emulator examples must be run from the flow-go-sdk/examples directory. Please see flow-go-sdk/examples/README.md for more details.")
+		} else {
+			fmt.Printf("Failed to load config from %s: %s\n", configPath, err.Error())
+		}
+
+		os.Exit(1)
+	}
 
 	d := json.NewDecoder(f)
 
--- a/examples/examples.go
+++ b/examples/examples.go
@@ -42,37 +42,47 @@ func ReadFile(path string) []byte {
 	return contents
 }
 
-var servicePrivateKeyHex string
+const configPath = "./flow.json"
 
-type flowJSON struct {
+var (
+	servicePrivateKeyHex     string
+	servicePrivateKeySigAlgo crypto.SignatureAlgorithm
+)
+
+type config struct {
 	Accounts struct {
 		Service struct {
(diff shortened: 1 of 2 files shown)


examples/flow.json
```diff
@@ -2,7 +2,7 @@
 	"accounts": {
 		"service": {
 			"address": "e467b9dd11fa00df",
-			"privateKey": "bf9db4706c2fdb9011ee7e170ccac492f05427b96ab41d8bf2d8c58443704b76",
+			"privateKey": "68ee617d9bf67a4677af80aaca5a090fcda80ff2f4dbc340e0e36201fa1f1d8c",
```
psiemens, line 5: `68ee617d9bf67a4677af80aaca5a090fcda80ff2f4dbc340e0e36201fa1f1d8c` is the default key for the emulator -- making it the same here so that users can run these examples against the emulator Docker image (which doesn't read from `flow.json`).

examples/examples.go
```diff
@@ -40,12 +42,49 @@ func ReadFile(path string) []byte {
...
+	var conf config
+
+	err = d.Decode(&conf)
+	Handle(err)
+
+	return conf
+}
+
+func init() {
+	conf := readConfig()
```
jordanschalm, line 78: The config path is relative, won't this fail for any program importing this package run from any other directory than `flow-go-sdk/examples`?

jordanschalm: For example the following program panics:
```go
package main

import (
    "fmt"
    _ "github.com/onflow/flow-go-sdk/examples"
)

func main() {
    fmt.Println("hello")
}
```

psiemens: Yeah, the examples are intended to be used from a directory that contains a `flow.json` file. I'd like for it to be a bit more flexible, but the `examples/README.md` encourages users to run both the emulator and the examples themselves from the `examples` directory.

The `examples` package isn't really meant to be imported by anything other than the examples in this directory. Do you think we should still make it a bit more robust?

psiemens: What about a friendlier error message? e.g.

> These examples must be run from a directory that contains a `flow.json` file.

jordanschalm: If it's documented in the README, I'm good with merging. The more descriptive is a good update too.

psiemens: Cool. Improved the error message here: https://github.com/onflow/flow-go-sdk/pull/49/commits/194382e3412be1d546752359e489d66cab0bfdd5

psiemens:
Thanks for the reviews @jordanschalm and @m4ksio!

PR #51, opened 2020-06-10
Update crypto comments

tarakby:
## Description
 - clarify why `InMemorySigner` is considered not secure (it is not simply because the key is in memory).
 - minor comment update

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -175,13 +175,16 @@ func (pk PublicKey) Encode() []byte {
 	return pk.publicKey.Encode()
 }
 
-// A Signer is capable of signing cryptographic messages.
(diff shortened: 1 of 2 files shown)


crypto/crypto.go
```diff
@@ -175,13 +175,16 @@ func (pk PublicKey) Encode() []byte {
...
+// A Signer is capable of generating cryptographic signatures.
 type Signer interface {
 	// Sign signs the given message with this signer.
 	Sign(message []byte) ([]byte, error)
 }
 
 // An InMemorySigner is a signer that generates signatures using an in-memory private key.
+//
+// InMemorySigner implements simple signing that does not protect the private key against
+// any tampering or side channel attacks.
```
psiemens, line 187: 💯

PR #52, opened 2020-06-10
Crypto: update secp256k1 implementation

tarakby:
## Description

Replace the current secp256k1 arithmetics by the faster bitcoin implementation.


crypto/internal/crypto/ecdsa_test.go
```diff
@@ -120,3 +121,82 @@ func TestECDSAUtils(t *testing.T) {
 		testKeySize(t, sk, ecdsaPrKeyLen[i], ecdsaPubKeyLen[i])
 	}
 }
+
+// TestScalarMult is a unit test of the scalar multiplication
+// This is only a sanity check meant to make sure the curve implemented
+// is checked against an independant test vector
```
psiemens, line 127:
```suggestion
// is checked against an independent test vector
```

PR #58, opened 2020-06-15
crypto: Remove KeyType enum

psiemens:
This enum is no longer being used can be removed from the `crypto` package.

We may revisit this API in the future and reintroduce a similar concept.

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -90,39 +90,20 @@ func StringToHashAlgorithm(s string) HashAlgorithm {
 	}
 }
 
-// KeyType is a key format supported by Flow.
-type KeyType int
-
-const (
-	UnknownKeyType KeyType = iota
-	ECDSA_P256_SHA2_256
-	ECDSA_P256_SHA3_256
-	ECDSA_secp256k1_SHA2_256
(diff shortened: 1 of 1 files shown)


crypto/crypto.go
```diff
@@ -90,39 +90,20 @@ func StringToHashAlgorithm(s string) HashAlgorithm {
...
-	switch k {
-	case ECDSA_P256_SHA2_256, ECDSA_secp256k1_SHA2_256:
-		return SHA2_256
-	case ECDSA_P256_SHA3_256, ECDSA_secp256k1_SHA3_256:
-		return SHA3_256
-	default:
-		return UnknownHashAlgorithm
+// CompatibleAlgorithms returns true if the signature and hash algorithms are compatible.
+func CompatibleAlgorithms(sigAlgo SignatureAlgorithm, hashAlgo HashAlgorithm) bool {
+	switch sigAlgo {
```
tarakby, line 95: It can also be written without branching:
```suggestion
		return (sigAlgo == ECDSA_P256 || sigAlgo == ECDSA_secp256k1) && (hashAlgo == SHA2_256 || hashAlgo == SHA3_256)
```

Issue #66, opened 2020-06-30
Transaction argument encoding is not consistent with other SDKs


psiemens:
### Problem

The `flow.Transaction` type holds arguments typed as `cadence.Value`, which means that the encoded form is not realized until the transaction is serialized for the wire or for hashing. This alone does not cause problems with the Go SDK, given that the Go's builtin JSON implementation is deterministic (the same arguments are encoded the same way every time).

However, if a user imports a transaction generated by the JavaScript SDK, decodes the arguments, and the constructs a `flow.Transaction` object, the final serialized format is not guaranteed to match the original transaction created in JS (i.e. differences in whitespace, indentation, etc.). This also means that the hashes will not match and all signatures will be invalid.

Given that JSON-CDC was not designed as a deterministic byte-wise encoding format, it does not make sense to expect the encoding to stay perfectly consistent across implementations. Thus, the Go SDK should represent transactions in a completely lossless way and avoid re-encoding arguments.

PR #67, opened 2020-07-01
Add RawArgument and Argument functions to flow.Transaction

psiemens:
Closes: #66 

## Description

This PR introduces a change to the way transaction arguments are stored on the `flow.Transaction` data type.

Because the _encoded_ arguments are used to determine the transaction hash, they must stay consistent across different representations of the same transaction. For example, when decoding a transaction from protobuf, we need to ensure that the original argument encoding is preserved in order to generate the same transaction hash.

Previously the `flow.Transaction` type was not lossless with regards to transaction arguments; by choosing to store arguments as `cadence.Value` types, we forced users of `flow.Transaction` to lose the original encoding (i.e. changes in JSON whitespace, indentation, etc).

This PR updates `flow.Transaction` to store arguments as raw bytes. 

- `AddArgument(arg)` can be used to add a `cadence.Value` argument
- `AddRawArgument(arg)` can be used to add a raw argument
- `Argument(i)` can be used to get an argument as a `cadence.Value`

### DIFF
--- a/client/convert/protobuf_test.go
+++ b/client/convert/protobuf_test.go
@@ -227,20 +227,6 @@ func TestConvert_Transaction(t *testing.T) {
 
 		assert.Equal(t, txA.ID(), txB.ID())
 	})
-
-	t.Run("With invalid arguments", func(t *testing.T) {
-		txA := test.TransactionGenerator().New()
-
-		msg, err := convert.TransactionToMessage(*txA)
-		require.NoError(t, err)
-
-		// invalid JSON-CDC
-		msg.Arguments = [][]byte{{1, 2, 3}}
-
-		txB, err := convert.MessageToTransaction(msg)
-		assert.Error(t, err)
-		assert.Equal(t, flow.Transaction{}, txB)
-	})
 }
 
 func TestConvert_TransactionResult(t *testing.T) {
--- a/transaction.go
+++ b/transaction.go
@@ -31,7 +31,7 @@ import (
 // A Transaction is a full transaction object containing a payload and signatures.
 type Transaction struct {
 	Script             []byte
-	Arguments          []cadence.Value
+	Arguments          [][]byte
 	ReferenceBlockID   Identifier
 	GasLimit           uint64
 	ProposalKey        ProposalKey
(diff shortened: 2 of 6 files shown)


transaction.go
```diff
@@ -59,10 +59,42 @@ func (t *Transaction) SetScript(script []byte) *Transaction {
 
 // AddArgument adds a Cadence argument to this transaction.
 func (t *Transaction) AddArgument(arg cadence.Value) *Transaction {
+	encodedArg, err := jsoncdc.Encode(arg)
+	if err != nil {
+		// argument is not added if encoding fails
```
Kay-Zee, line 64: no failure signal is returned here. Is this to keep the interface backward compatible?

client/convert/protobuf_test.go
```diff
@@ -227,20 +227,6 @@ func TestConvert_Transaction(t *testing.T) {
 
 		assert.Equal(t, txA.ID(), txB.ID())
 	})
-
-	t.Run("With invalid arguments", func(t *testing.T) {
```
Kay-Zee, line 231: this test no longer possible because we accept arbitrary bytes?

transaction.go
psiemens: Yeah, that was my original intent, but I'm having second thoughts. I think this should be more explicit about the possibility of an error.

@Kay-Zee @turbolent which API seems best to you? I leaning towards option D. Even though its probably the worst UX, it's by far the most explicit. 

**Option A**
```go
var myArg cadence.Value

tx.AddArgument(myArg) // no error, fails silently

tx.AddRawArgument(jsoncdc.MustEncode(myArg))
```

**Option B**
```go
var myArg cadence.Value

tx.AddArgument(myArg) // panics if encoding fails

tx.AddRawArgument(jsoncdc.MustEncode(myArg))
```

**Option C**
```go
var myArg cadence.Value

err := tx.AddArgument(myArg) // error for failed encoding, no chainability

tx.AddRawArgument(jsoncdc.MustEncode(myArg))
```

**Option D**

Caller handles all encoding

```go
var myArg cadence.Value

// AddArgument accepts []byte, encoding must be done by caller
// Note: AddRawArgument(b []byte) and Argument(i int) functions removed
tx.AddArgument(jsoncdc.MustEncode(myArg))
```

Kay-Zee: I'm actually more partial to option B, with a name change:

```go
var myArg cadence.Value

tx.MustAddArgumentValue(myArg) // panics if encoding fails

tx.AddArgument(jsoncdc.MustEncode(myArg))
```

with `MustAddArgumentValue` being just a slice convenience of not having to import `jsoncdc` for the `MustEncode` function. The naming change will also make having the user handle the encoding seem more like the default here, which might actually be what we want.

I'm overall most worried about failing silently, since i know it's not really suppose to happen, but still...

client/convert/protobuf_test.go
psiemens: Yep, Cadence encoding can't fail here anymore

transaction.go
psiemens: Another thing I just thought of is the fact that we may want to swap the argument encoding format in the future, and therefor it could be good to keep JSON-CDC completely separate from the `flow.Transaction` implementation.

Seeing as your proposal builds on top of option D, how would you feel about merging option D today and then adding the `MustAddArgumentValue` as a convenience function later if there ends up being demand for it?

turbolent: I think it's very likely and desirable that we use a more efficient binary encoding format for the arguments soon, so it might be a good idea to keep it flexible. However, this would be additive and we would need a way to differentiate (tag) the encoded data.

I don't really see a problem of having the SDK know about supported encoding formats and providing the convenience of performing the best encoding given a `cadence.Value` (one could even imagine it's based on a "debug mode" flag that uses the more human-readable JSON format instead of the more efficient binary format)

psiemens: After giving this more thought and trying out some sample code, I think I like option C the best. Although the error return breaks chain-ability of the setter method, I think it's still the best combination of correctness (error is not implicit or hidden) and convenience for the users.

@Kay-Zee @turbolent Do you mind taking another look?

transaction.go
```diff
@@ -58,11 +58,42 @@ func (t *Transaction) SetScript(script []byte) *Transaction {
...
+	encodedArg, err := jsoncdc.Encode(arg)
+	if err != nil {
+		return fmt.Errorf("failed to encode argument: %w", err)
+	}
+
+	t.Arguments = append(t.Arguments, encodedArg)
+	return nil
+}
+
+// AddRawArgument adds a raw JSON-CDC encoded argument to this transaction.
```
turbolent, line 71: Maybe this comment shouldn't mention JSON-CDC, as it itself does not require the passed value to be encoded that way. In the future this might confuse readers/users: it sounds like that this function only accepts JSON-CDC, though in the future it would/could be used with another encoding format.

PR #78, opened 2020-07-07
Update account management templates to use parameterized transactions

psiemens:
## Description

This PR updates the account management templates in `templates` to use parameterized transactions rather than string templating.

The template functions now return partial `flow.Transaction` objects, which can later be signed and sent to the network.


templates/accounts.go
```diff
@@ -19,113 +19,109 @@
...
+  }
+}
+`
 
-	script := fmt.Sprintf(
-		`
-            transaction {
-              prepare(signer: AuthAccount) {
-                let acct = AuthAccount(payer: signer)
+// CreateAccount generates a transactions that creates a new account.
```
turbolent, line 42: 👍

PR #91, opened 2020-08-26
Document the transaction fields

turbolent:
It was not clear what the block reference field means. I copied variants of the comments from https://github.com/dapperlabs/flow-go/blob/25a69c092fca3efe3f2c140c5d0acbff4e391b3e/model/flow/transaction.go


transaction.go
```diff
@@ -30,14 +30,35 @@ import (
...
+	// This reference can be set to an older block if the tx should expire faster.
+	ReferenceBlockID Identifier
+
+	// GasLimit is the max amount of computation which is allowed to be used during this transaction
+	GasLimit uint64
+
+	// ProposalKey is key of the account used to propose the transaction
+	ProposalKey ProposalKey
+
+	// Payer is the account that pays the feed for this transaction
```
joshuahannan, line 53:
```suggestion
	// Payer is the account that pays the fee for this transaction
```

transaction.go
```diff
@@ -30,14 +30,35 @@ import (
...
+	Payer Address
+
+	Authorizers []Address
+
+	// PayloadSignatures is the list of account signatures excluding signature of the payer account
+	PayloadSignatures []TransactionSignature
+
+	// EnvelopeSignatures is the payer signature over the envelope (payload + payload signatures)
```
joshuahannan, line 61: Could you clarify this a bit? Should it be `payer + payload` signatures?

cybercent:
One field that needs a bit of clarification is `Authorizers`

transaction.go
turbolent: Sorry, I just copied this over from the flow-go codebase

joshuahannan: I think I understand what it means now. I thought it was saying (payload signature + payload signature) and I thought it was a typo. But is actually saying (the entire payload + the signature for the payload).

turbolent:
@psiemens would you be able to add a comment for `Authorizers`? the one in flow-go isn't really worth copying over

psiemens:
@turbolent I updated the comments (8c4d958) to add a bit more clarity -- I found that the ones from `flow-go` weren't quite descriptive enough for people unfamiliar with Flow internals. I also added links to our new docs site for those who want to read more.

turbolent:
Nice, so much better, looks good! Needs an approval from an owner and then I can merge.

psiemens (approved):
🚢

PR #96, opened 2020-09-15
crypto/cloudkms: Add initial draft of Google Cloud KMS integration

psiemens:
This PR adds initial support for using Google Cloud KMS to sign Flow transactions.

`cloudkms.Signer` implements `crypto.Signer` and can be used as a drop-in replacement for `crypto.InMemorySigner`


psiemens:
@Kay-Zee @cybercent I updated this PR with a function to fetch public keys from KMS and convert them into `crypto.PublicKey` values.

Also, the `cloudkms.Signer` now automatically instantiates a hasher based on the algorithm it fetches with `GetPublicKey`

crypto/cloudkms/cloudkms.go
```diff
@@ -0,0 +1,251 @@
...
+		return crypto.PublicKey{},
+			crypto.UnknownHashAlgorithm,
+			fmt.Errorf("cryptokms: failed to parse PEM public key: %w", err)
+	}
+
+	return publicKey, hashAlgo, nil
+}
+
+// Signer is a Google Cloud KMS implementation of crypto.Signer.
+type Signer struct {
```
Kay-Zee, line 129: This file is still pretty compact, so this could be a bit premature, but could consider splitting out the file based on the structs.

Kay-Zee (approved):
LGTM

Kay-Zee (approved):

PR #97, opened 2020-09-16
templates: Do not deploy empty code string on account creation

psiemens:
This fixes an inefficiency in the account creation template that was causing a the `setCode` method to always be called, even if there is no code to deploy.


turbolent (approved):
Good catch!

Issue #99, opened 2020-09-21
Add BlockTimestamp to BlockEvents


janezpodhostnik:
### Description

Add BlockTimestamp to BlockEvents so you can order the events into the timeline if you need to.

ref: https://github.com/onflow/flow/issues/40
ref: https://github.com/onflow/flow/pull/41

### Acceptance Criteria

- [ ] Added event `BlockTimestamp` field to `BlockEvents`

PR #100, opened 2020-09-21
Added BlockTimestamp to BlockEvents response

janezpodhostnik:
Closes: #99

## Description

Added BlockTimestamp to the access node EventsResponse. This allows you to know when the event was published without re-requesting the block by block id.


psiemens (approved):
LGTM 🚢

PR #103, opened 2020-09-30
Updated sdk examples

janezpodhostnik:
## Description

SDK examples didn't set ReferenceBlockId which has become mandatory. Emulator was updated to also require this: https://github.com/dapperlabs/flow-emulator/pull/103

### DIFF
--- a/examples/create_account/main.go
+++ b/examples/create_account/main.go
@@ -48,14 +48,14 @@ func CreateAccountDemo() {
 		SetHashAlgo(crypto.SHA3_256).
 		SetWeight(flow.AccountKeyWeightThreshold)
 
-	referenceBlockId := examples.GetReferenceBlockId(flowClient)
+	referenceBlockID := examples.GetReferenceBlockId(flowClient)
 	createAccountTx := templates.CreateAccount([]*flow.AccountKey{myAcctKey}, nil, serviceAcctAddr)
 	createAccountTx.SetProposalKey(
 		serviceAcctAddr,
 		serviceAcctKey.Index,
 		serviceAcctKey.SequenceNumber,
 	)
-	createAccountTx.SetReferenceBlockID(referenceBlockId)
+	createAccountTx.SetReferenceBlockID(referenceBlockID)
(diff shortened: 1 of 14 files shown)


examples/examples.go
```diff
@@ -141,6 +141,12 @@ func RandomAccount(flowClient *client.Client) (flow.Address, *flow.AccountKey, c
 	return account.Address, account.Keys[0], signer
 }
 
+func GetReferenceBlockId(flowClient *client.Client) flow.Identifier {
+	blk, err := flowClient.GetLatestBlock(context.Background(), false)
```
psiemens, line 145: It's probably better to set `isSealed=true` here:
```suggestion
	block, err := flowClient.GetLatestBlock(context.Background(), true)
```

examples/examples.go
```diff
@@ -141,6 +141,12 @@ func RandomAccount(flowClient *client.Client) (flow.Address, *flow.AccountKey, c
...
+	Handle(err)
+	return blk.ID
```
psiemens, line 147:
```suggestion
	return block.ID
```

examples/create_account/main.go
```diff
@@ -48,12 +48,14 @@ func CreateAccountDemo() {
 		SetHashAlgo(crypto.SHA3_256).
 		SetWeight(flow.AccountKeyWeightThreshold)
 
+	referenceBlockId := examples.GetReferenceBlockId(flowClient)
```
psiemens, line 51:
```suggestion
	referenceBlockID := examples.GetReferenceBlockId(flowClient)
```

Go convention is to always capitalized acronyms/abbreviations:

turbolent:
@janezpodhostnik Looks good! it looks like @Kay-Zee already updated dependencies in #103, so that's why there's a conflict in the go.mod file.

Not quite sure, what is the reason for this also adding Google Cloud KMS code?

turbolent (approved):
Looks good, nice!

psiemens (approved):

turbolent (approved):
👍

PR #106, opened 2020-10-05
Multiple smart contract support

janezpodhostnik:
References: https://github.com/dapperlabs/flow-go/issues/4806

## Description

______

For contributor use:

### DIFF
--- a/account.go
+++ b/account.go
@@ -26,10 +26,11 @@ import (
 
 // An Account is an account on the Flow network.
 type Account struct {
-	Address Address
-	Balance uint64
-	Code    []byte
-	Keys    []*AccountKey
+	Address   Address
+	Balance   uint64
+	Code      []byte
+	Keys      []*AccountKey
+	Contracts map[string][]byte
 }
 
 // AccountKeyWeightThreshold is the total key weight required to authorize access to an account.
--- a/client/convert/protobuf.go
+++ b/client/convert/protobuf.go
@@ -42,10 +42,11 @@ func AccountToMessage(a flow.Account) *entities.Account {
 	}
 
 	return &entities.Account{
-		Address: a.Address.Bytes(),
-		Balance: a.Balance,
-		Code:    a.Code,
-		Keys:    accountKeys,
+		Address:   a.Address.Bytes(),
+		Balance:   a.Balance,
+		Code:      a.Code,
+		Keys:      accountKeys,
+		Contracts: a.Contracts,
 	}
 }
 
@@ -65,10 +66,11 @@ func MessageToAccount(m *entities.Account) (flow.Account, error) {
 	}
 
 	return flow.Account{
-		Address: flow.BytesToAddress(m.GetAddress()),
-		Balance: m.GetBalance(),
-		Code:    m.GetCode(),
-		Keys:    accountKeys,
+		Address:   flow.BytesToAddress(m.GetAddress()),
+		Balance:   m.GetBalance(),
+		Code:      m.GetCode(),
+		Keys:      accountKeys,
+		Contracts: m.GetContracts(),
 	}, nil
 }
 
--- a/examples/query_events/main.go
+++ b/examples/query_events/main.go
@@ -21,6 +21,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/onflow/flow-go-sdk/templates"
 
 	"google.golang.org/grpc"
 
@@ -53,7 +54,11 @@ func QueryEventsDemo() {
 		}
 	`
 
-	contractAddr := examples.DeployContract(flowClient, []byte(contract))
+	contractAccount := examples.CreateAccountWithContracts(flowClient,
+		nil, []templates.Contract{{
+			Name:   "EventDemo",
+			Source: contract,
+		}})
 
 	// Send a tx that emits the event in the deployed contract
 	script := fmt.Sprintf(`
@@ -64,7 +69,7 @@ func QueryEventsDemo() {
 				EventDemo.add(x: 2, y: 3)
 			}
 		}
-	`, contractAddr.Hex())
+	`, contractAccount.Address.Hex())
 
 	referenceBlockID := examples.GetReferenceBlockId(flowClient)
 	runScriptTx := flow.NewTransaction().
@@ -103,7 +108,7 @@ func QueryEventsDemo() {
 	// 2
 	// Query for our custom event by type
 	results, err = flowClient.GetEventsForHeightRange(ctx, client.EventRangeQuery{
-		Type:        fmt.Sprintf("A.%s.EventDemo.Add", contractAddr.Hex()),
+		Type:        fmt.Sprintf("AC.%s.EventDemo.EventDemo.Add", contractAccount.Address.Hex()),
 		StartHeight: 0,
 		EndHeight:   100,
 	})
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,8 @@ require (
 	github.com/btcsuite/btcd v0.0.0-20171128150713-2e60448ffcc6
 	github.com/ethereum/go-ethereum v1.9.9
 	github.com/golang/protobuf v1.4.2
-	github.com/onflow/cadence v0.9.1
-	github.com/onflow/flow/protobuf/go/flow v0.1.7
+	github.com/onflow/cadence v0.10.0
+	github.com/onflow/flow/protobuf/go/flow v0.1.8
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/objx v0.1.1 // indirect
 	github.com/stretchr/testify v1.5.1
@@ -17,3 +17,5 @@ require (
 	google.golang.org/genproto v0.0.0-20200831141814-d751682dd103
 	google.golang.org/grpc v1.31.1
 )
+
+replace github.com/fxamacker/cbor/v2 => github.com/turbolent/cbor/v2 v2.2.1-0.20200911003300-cac23af49154
--- a/templates/accounts.go
+++ b/templates/accounts.go
@@ -25,68 +25,117 @@ import (
 	"github.com/onflow/flow-go-sdk"
 )
 
-const createAccountTemplate = `
-transaction(publicKeys: [[UInt8]], code: [UInt8]) {
-  prepare(signer: AuthAccount) {
-	let acct = AuthAccount(payer: signer)
+// Contract is a Cadence contract deployed to a Flow account.
+//
+// Name is the identifier for the contract within the account.
+//
+// Source is the the Cadence code of the contract.
+type Contract struct {
+	Name   string
+	Source string
+}
 
-	for key in publicKeys {
-	  acct.addPublicKey(key)
+// SourceBytes returns the UTF-8 encoded source code (Source) of the contract.
+func (c Contract) SourceBytes() []byte {
+	return []byte(c.Source)
+}
+
+const createAccountTemplate = `
+transaction(publicKeys: [[UInt8]], contracts: {String: [UInt8]}) {
+	prepare(signer: AuthAccount) {
+		let acct = AuthAccount(payer: signer)
+
+		for key in publicKeys {
+			acct.addPublicKey(key)
+		}
+
+		for contract in contracts.keys {
+			acct.contracts.add(name: contract, code: contracts[contract]!)
+		}
 	}
-	
-	if code.length > 0 {
-	  acct.setCode(code)
-	}
-  }
 }
 `
 
 // CreateAccount generates a transactions that creates a new account.
 //
-// This template accepts a list of public keys and a code argument, both of which are optional.
+// This template accepts a list of public keys and a contracts argument, both of which are optional.
+//
+// The contracts argument is a dictionary of *contract name*: *contract code (in bytes)*.
+// All of the contracts will be deployed to the account.
 //
 // The final argument is the address of the account that will pay the account creation fee.
 // This account is added as a transaction authorizer and therefore must sign the resulting transaction.
-func CreateAccount(accountKeys []*flow.AccountKey, code []byte, payer flow.Address) *flow.Transaction {
+func CreateAccount(accountKeys []*flow.AccountKey, contracts []Contract, payer flow.Address) *flow.Transaction {
 	publicKeys := make([]cadence.Value, len(accountKeys))
 
 	for i, accountKey := range accountKeys {
 		publicKeys[i] = bytesToCadenceArray(accountKey.Encode())
 	}
 
+	contractKeyPairs := make([]cadence.KeyValuePair, len(contracts))
+
+	for i, contract := range contracts {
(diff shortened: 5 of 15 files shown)


go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	github.com/btcsuite/btcd v0.0.0-20171128150713-2e60448ffcc6
 	github.com/ethereum/go-ethereum v1.9.9
 	github.com/golang/protobuf v1.4.2
-	github.com/onflow/cadence v0.9.1
-	github.com/onflow/flow/protobuf/go/flow v0.1.7
+	github.com/onflow/cadence v0.10.0
+	github.com/onflow/flow/protobuf/go/flow v0.1.8-0.20201002204652-a7175c92529c
```
turbolent, line 11: We should never use commits of branches, because they get deleted.

Commit ca151b05acef0c2d802f92bc5f2a15b663c17c67 in https://github.com/onflow/flow should get tagged and referenced here

go.mod
```diff
@@ -17,3 +17,5 @@ require (
 	google.golang.org/genproto v0.0.0-20200831141814-d751682dd103
 	google.golang.org/grpc v1.31.1
 )
+
+replace github.com/fxamacker/cbor/v2 => github.com/turbolent/cbor/v2 v2.2.1-0.20200911003300-cac23af49154
```
turbolent, line 21: Good catch! I'm pretty close to getting this upstreamed, so hopefully we can remove it again soon

examples/query_events/main.go
```diff
@@ -103,7 +104,7 @@ func QueryEventsDemo() {
 	// 2
 	// Query for our custom event by type
 	results, err = flowClient.GetEventsForHeightRange(ctx, client.EventRangeQuery{
-		Type:        fmt.Sprintf("A.%s.EventDemo.Add", contractAddr.Hex()),
+		Type:        fmt.Sprintf("A.%s.EventDemo.Add", contractAccount.Address.Hex()),
```
turbolent, line 107: I think the event type should now be different here, no?

account.go
```diff
@@ -26,10 +26,11 @@ import (
...
 type Account struct {
-	Address Address
-	Balance uint64
-	Code    []byte
-	Keys    []*AccountKey
+	Address   Address
+	Balance   uint64
+	Code      []byte
+	Keys      []*AccountKey
+	Contracts map[string][]byte
```
turbolent, line 33: 👌

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
 	  acct.addPublicKey(key)
 	}
-	
-	if code.length > 0 {
-	  acct.setCode(code)
+  }
+}
+`
+
+const createAccountWitContractsTemplate = `
```
turbolent, line 40: 👍

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
+transaction(publicKeys: [[UInt8]], contracts: {String: [UInt8]}) {
+  prepare(signer: AuthAccount) {
+	let acct = AuthAccount(payer: signer)
+
+	for key in publicKeys {
+	  acct.addPublicKey(key)
+	}
+
+    for contract in contracts.keys {
+	  acct.contracts.add(name: contract, code: contracts[contract]!)
```
turbolent, line 50: Would be nice to add key-value iteration over dictionaries to make this more ergonomic

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
-		AddRawArgument(jsoncdc.MustEncode(cadenceCode))
+		AddRawArgument(jsoncdc.MustEncode(cadenceContracts))
 }
 
 const updateAccountCodeTemplate = `
-transaction(code: [UInt8]) {
+transaction(name: String, code: [UInt8]) {
   prepare(signer: AuthAccount) {
-	signer.setCode(code)
+	signer.contracts.update__experimental(name: name, code: code)
```
turbolent, line 116: 👍

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
 	cadenceCode := bytesToCadenceArray(code)
 
 	return flow.NewTransaction().
 		SetScript([]byte(updateAccountCodeTemplate)).
+		AddRawArgument(jsoncdc.MustEncode(cadenceName)).
+		AddRawArgument(jsoncdc.MustEncode(cadenceCode)).
+		AddAuthorizer(address)
+}
+
+const setAccountCodeTemplate = `
```
turbolent, line 133: Maybe rename "set" to "add"

turbolent (commented):
Looks good 👍 Just a couple minor points

templates/accounts.go
janezpodhostnik: There should also be an example of iterating over a dictionary in the docs. I can add that. :)

turbolent: Oh, this was more like an observation/note for myself! Cadence doesn't support that yet, it would be nice to add it there 😅

janezpodhostnik: What I meant is that it took me a bit to figure this out 😄 :
```
    for contract in contracts.keys {
	  acct.contracts.add(name: contract, code: contracts[contract]!)
    }
```

As it is not mentioned in the docs right now.

turbolent: Oh, I misunderstood! Good point, I added an example to the documentation here: https://github.com/onflow/cadence/pull/399

go.mod
turbolent: Tagged that repo and updated the dependency in 882d2e0

templates/accounts.go
psiemens, line 50: Indentation is a bit inconsistent in this block

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
+
+	return flow.NewTransaction().
+		SetScript([]byte(createAccountWitContractsTemplate)).
+		AddAuthorizer(payer).
 		AddRawArgument(jsoncdc.MustEncode(cadencePublicKeys)).
-		AddRawArgument(jsoncdc.MustEncode(cadenceCode))
+		AddRawArgument(jsoncdc.MustEncode(cadenceContracts))
 }
 
 const updateAccountCodeTemplate = `
```
psiemens, line 113:
```suggestion
const updateAccountContractTemplate = `
```

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
   prepare(signer: AuthAccount) {
-	signer.setCode(code)
+	signer.contracts.update__experimental(name: name, code: code)
   }
 }
 `
 
 // UpdateAccountCode generates a transaction that updates the code deployed at an account.
-func UpdateAccountCode(address flow.Address, code []byte) *flow.Transaction {
+func UpdateAccountCode(address flow.Address, name string, code []byte) *flow.Transaction {
```
psiemens, line 122:
```suggestion
func UpdateAccountContract(address flow.Address, name string, code []byte) *flow.Transaction {
```

In general I like the idea of retiring the term "code" in our SDKs -- I think it makes a lot more sense to refer to specific contracts on an account.

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
+const addAccountCodeTemplate = `
+transaction(name: String, code: [UInt8]) {
+  prepare(signer: AuthAccount) {
+	signer.contracts.add(name: name, code: code)
+  }
+}
+`
+
+// SetAccountCode generates a transaction that sets the code deployed at an account.
+func SetAccountCode(address flow.Address, name string, code []byte) *flow.Transaction {
```
psiemens, line 142:
```suggestion
func SetAccountContract(address flow.Address, name string, code []byte) *flow.Transaction {
```

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
+// CreateAccountWithContracts generates a transactions that creates a new account.
+//
+// This template accepts a list of public keys and a contracts argument, both of which are optional.
+//
+// The contracts argument is a dictionary of *contract name*: *contract code (in bytes)*.
+// All of the contracts will be deployed to the account.
+//
+// The final argument is the address of the account that will pay the account creation fee.
+// This account is added as a transaction authorizer and therefore must sign the resulting transaction.
+func CreateAccountWithContracts(accountKeys []*flow.AccountKey, contracts map[string][]byte, payer flow.Address) *flow.Transaction {
```
psiemens, line 86: Do you it's worthwhile to have a separate function for this? I'm honestly not sure, but it would be possible to just modify `CreateAccount` and make the contracts argument optional. 

```go
func CreateAccount(accountKeys []*flow.AccountKey, contracts map[string][]byte, payer flow.Address)  { }
```

If `contracts` is `nil`, the loop below will just behave as if `contracts` is an empty map

templates/accounts.go
```diff
@@ -26,58 +26,126 @@ import (
...
+	publicKeys := make([]cadence.Value, len(accountKeys))
+
+	for i, accountKey := range accountKeys {
+		publicKeys[i] = bytesToCadenceArray(accountKey.Encode())
+	}
+
+	contractKeyPairs := make([]cadence.KeyValuePair, len(contracts))
+	i := 0
+	for s, bytes := range contracts {
```
psiemens, line 95: Map iteration order in Go is non-deterministic, so this function may generate different Cadence argument orderings on subsequent runs with the same input.

This probably isn't a problem in most cases, but could be confusing for users who expect deterministic transaction generation in their tests. 

What if we added a dedicated type for contracts? It could even be defined in the `cadence` package:

```go
type Contract struct {
  Name string
  Source []byte
}
```

```go
import "github.com/onflow/cadence"

func CreateAccountWithContracts(accountKeys []*flow.AccountKey, contracts []cadence.Contract, payer flow.Address) *flow.Transaction {
  // ... 
}

contracts := []cadence.Contract{
  cadence.NewContract("MyContract", myContractCode),
}

tx := CreateAccountWithContracts(keys, contracts, payer)
```

@turbolent @janezpodhostnik

psiemens (commented):
This is a great addition!

templates/accounts.go
turbolent: Good point Peter! The recommendation looks good

turbolent (approved):
Nice work! 👏

templates/accounts.go
janezpodhostnik: @turbolent @psiemens switched to list. I double checked there isn't a similar issue in flow-go.

turbolent:
@janezpodhostnik 8d0c6ea07197b3c1ecec5ee31900e97e26b74fdc looks good! 🚀

templates/accounts.go
```diff
@@ -26,67 +26,118 @@ import (
...
+	}
 }
+`
 
-const updateAccountCodeTemplate = `
-transaction(code: [UInt8]) {
-  prepare(signer: AuthAccount) {
-	signer.setCode(code)
-  }
+// UpdateAccountContract generates a transaction that updates the contract deployed at an account.
```
psiemens, line 104:
```suggestion
// UpdateAccountContract generates a transaction that updates a contract deployed at an account.
```

Current phrasing implies only one contract per account.

templates/accounts.go
```diff
@@ -26,67 +26,118 @@ import (
...
 	return flow.NewTransaction().
-		SetScript([]byte(createAccountTemplate)).
+		SetScript([]byte(createAccountWitContractsTemplate)).
 		AddAuthorizer(payer).
 		AddRawArgument(jsoncdc.MustEncode(cadencePublicKeys)).
-		AddRawArgument(jsoncdc.MustEncode(cadenceCode))
+		AddRawArgument(jsoncdc.MustEncode(cadenceContracts))
+}
+
+type Contract struct {
```
psiemens, line 91: Can you add a comment to this? 

Something like:

```
// Contract is a Cadence contract deployed to a Flow account.
//
// Name is the identifier for the contract within the account.
//
// Source is the UTF-8 encoded source code for the contract.
```

I would also define this struct before the `CreateAccount` function

psiemens: @turbolent @janezpodhostnik Do you think that `Source` should just be a string? We originally chose bytes so that it'd be easier to swap this out for compiled code (e.g. WASM) in the future, but I'm wondering if this specific struct can just accept a string for convenience. 

The name `Source` does imply that it is source code, which to me also implies that it is a human readable string. Perhaps there could be a separate `Compiled` field in the future.

templates/accounts.go
```diff
@@ -26,67 +26,118 @@ import (
...
-	  acct.addPublicKey(key)
+		for key in publicKeys {
+			acct.addPublicKey(key)
+		}
 	}
-	
-	if code.length > 0 {
-	  acct.setCode(code)
+}
+`
```
psiemens, line 38: This template isn't used anymore, right? I would rename the template below to `createAccountTemplate` to avoid confusion, and also to show that it is used by the `CreateAccount` function.

client/convert/protobuf.go
```diff
@@ -65,10 +66,11 @@ func MessageToAccount(m *entities.Account) (flow.Account, error) {
...
 	return flow.Account{
-		Address: flow.BytesToAddress(m.GetAddress()),
-		Balance: m.GetBalance(),
-		Code:    m.GetCode(),
-		Keys:    accountKeys,
+		Address:   flow.BytesToAddress(m.GetAddress()),
+		Balance:   m.GetBalance(),
+		Code:      m.GetCode(),
+		Keys:      accountKeys,
+		Contracts: m.GetContracts(),
```
psiemens, line 73: Seeing this again, I'm now wondering if the contract ordering matters in the protobuf response, too. 

The order of keys in a serialized protobuf map is undefined, so we can't be sure that this `Contracts` field will always be returned in the same order. But is that okay?

templates/accounts.go
janezpodhostnik: I agree Source would be easier to use if it was a string. I also added a GetSourceBytes function to the Contract for ease of use.

templates/accounts.go
janezpodhostnik: sorry didn't catch that wasn't used anymore. I applied the suggested changes.

client/convert/protobuf.go
janezpodhostnik: When deploying the contracts order matters to have a consistent state commitment. When getting the contracts I don't se a reason why the order should be deterministic. How does the order matter to the user?

On the other hand its slightly weird that you put an array in and get out a map...

templates/accounts.go
turbolent: It's bytes in the protobuf and state, but I think it's convenient that it's a high-level string in the SDK.

client/convert/protobuf.go
turbolent: Hmm, it shouldn't matter, should it? I guess if one wanted to reconstruct, one couldn't. 
Given that the `contracts` field is in the protobuf, but was only just added, we can still change it easily?

psiemens (approved):
LGTM, just one more small suggestion!

I think we should keep the protobuf format as-is 👍

PR #107, opened 2020-11-02
[BUGFIX] fix readme

qq976739120:
It looks like you haven't updated it in a long time.I've only fixed part of the problem,And make it more in line with the GO specification.I would like to work for this SDK for a long time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,71 +96,83 @@ Once you have [generated a key pair](#generating-keys), you can create a new acc
 using its public key.
 
 ```go
+package main
+
 import (
-    "github.com/onflow/flow-go-sdk"
-    "github.com/onflow/flow-go-sdk/crypto"
-    "github.com/onflow/flow-go-sdk/templates"
+	"context"
+	"fmt"
+
+	"google.golang.org/grpc"
+
+	"github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go-sdk/client"
+	"github.com/onflow/flow-go-sdk/crypto"
+	"github.com/onflow/flow-go-sdk/examples"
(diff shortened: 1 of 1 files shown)


psiemens (commented):
These are great suggestions, thank you @qq976739120!

I originally wrote the examples as partial Go programs in order to keep them short. Seeing them now, I agree that it'd be better if they would actually compile. 

Do you think it would make more sense to remove the code snippets from the README and instead just link directly to the examples in the `/examples` directory?

qq976739120:
> These are great suggestions, thank you @qq976739120!
> 
> I originally wrote the examples as partial Go programs in order to keep them short. Seeing them now, I agree that it'd be better if they would actually compile.
> 
> Do you think it would make more sense to remove the code snippets from the README and instead just link directly to the examples in the `/examples` directory?
This seems like a big change.I looked carefully at the code  in the `/examples` directory,I think I have some optimizations.

qq976739120:
> Do you think it would make more sense to remove the code snippets from the README and instead just link directly to the examples in the `/examples` directory?

This seems like a big change.I looked carefully at the code in the /examples directory,I think I have some optimizations.

psiemens:
> I looked carefully at the code in the /examples directory, I think I have some optimizations.

I agree. Is that something you'd want to work on? We're very open to contributions 😄

qq976739120:
> I agree. Is that something you'd want to work on? We're very open to contributions 😄

Is there a group of developers to communicate with? More detailed communication is needed. I don't have it on the official website

psiemens:
@qq976739120 You can find us in our Discord server: https://discord.gg/flow

We mostly talk about this project in the `#flow-go-sdk` channel. Happy to chat there!

PR #111, opened 2020-11-11
Fix transaction size growing very quickly

janezpodhostnik:
ref: https://github.com/dapperlabs/flow-go/issues/5056

## Description

Using `[Uint8]` parameter type to pass contract code to a transaction was causing the transaction size to grow with 30 bytes per byte of contract code. 

This was because converting a byte array to a cadence Uint array wrapped every byte with a lot extra type data.

I circumvented this by passing the code to the transaction as a hex encoded string instead of a `[Uint8]`. Doing it this way every byte of contract code increases the transaction size by only 2 bytes (because of the hex encoding).

A possible future expansion would be to use base64 encoding which would have a ratio of 1.3. Or even better than that, some sort of compression.

This was tested with this emulator version https://github.com/onflow/flow-emulator/tree/bastian/update-sdk-v0.12.0
______

For contributor use:

### DIFF
--- a/templates/accounts.go
+++ b/templates/accounts.go
@@ -36,18 +36,23 @@ type Contract struct {
 	Source string
 }
 
+// SourceBytes returns the UTF-8 encoded source code (Source) of the contract.
+func (c Contract) SourceBytes() []byte {
+	return []byte(c.Source)
+}
+
 // SourceHex returns the UTF-8 encoded source code (Source) of the contract as a hex string.
 func (c Contract) SourceHex() string {
-	return hex.EncodeToString([]byte(c.Source))
+	return hex.EncodeToString(c.SourceBytes())
 }
 
 const createAccountTemplate = `
-transaction(publicKeys: [[UInt8]], contracts: {String: String}) {
+transaction(publicKeys: [String], contracts: {String: String}) {
 	prepare(signer: AuthAccount) {
 		let acct = AuthAccount(payer: signer)
 
 		for key in publicKeys {
-			acct.addPublicKey(key)
+			acct.addPublicKey(key.decodeHex())
 		}
 
 		for contract in contracts.keys {
@@ -70,7 +75,8 @@ func CreateAccount(accountKeys []*flow.AccountKey, contracts []Contract, payer f
 	publicKeys := make([]cadence.Value, len(accountKeys))
 
 	for i, accountKey := range accountKeys {
-		publicKeys[i] = bytesToCadenceArray(accountKey.Encode())
+		keyHex := hex.EncodeToString(accountKey.Encode())
+		publicKeys[i] = cadence.NewString(keyHex)
 	}
 
 	contractKeyPairs := make([]cadence.KeyValuePair, len(contracts))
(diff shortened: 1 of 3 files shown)


templates/accounts.go
```diff
@@ -35,13 +36,13 @@ type Contract struct {
...
-func (c Contract) SourceBytes() []byte {
-	return []byte(c.Source)
+// SourceHex returns the UTF-8 encoded source code (Source) of the contract as a hex string.
+func (c Contract) SourceHex() string {
+	return hex.EncodeToString([]byte(c.Source))
 }
 
 const createAccountTemplate = `
-transaction(publicKeys: [[UInt8]], contracts: {String: [UInt8]}) {
+transaction(publicKeys: [[UInt8]], contracts: {String: String}) {
```
cybercent, line 45: Are the `publicKeys: [[UInt8]]` also increased 30 bytes per byte?

templates/accounts.go
```diff
@@ -35,13 +36,13 @@ type Contract struct {
 	Source string
 }
 
-// SourceBytes returns the UTF-8 encoded source code (Source) of the contract.
-func (c Contract) SourceBytes() []byte {
-	return []byte(c.Source)
+// SourceHex returns the UTF-8 encoded source code (Source) of the contract as a hex string.
```
psiemens, line 39: I think we might as well keep the `SourceBytes` method

templates/accounts_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go-sdk/templates"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestCreateAccount(t *testing.T) {
+	// Converting transaction arguments to cadence values, can increase their size.
```
psiemens, line 29:
```suggestion
	// Converting transaction arguments to Cadence values can increase their size.
```

templates/accounts_test.go
```diff
@@ -0,0 +1,53 @@
...
+	// If this is not taken into account the transaction can quickly grow over the maximum transaction size limit.
+	t.Run("Transaction should not grow uncontrollably in size", func(t *testing.T) {
```
psiemens, line 31: Great test!

psiemens (approved):
Great optimization!

turbolent:
Nice, this is what we used to do before 👍 

Maybe we can also apply the same optimization to public keys, like @cybercent pointed out?

turbolent (approved):

templates/accounts.go
janezpodhostnik: Good catch!

janezpodhostnik:
I implemented the suggested changes.

Is there documentation related to this I need to change?

templates/accounts_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestCreateAccount(t *testing.T) {
+	// Converting transaction arguments to Cadence values can increase their size.
+	// If this is not taken into account the transaction can quickly grow over the maximum transaction size limit.
+	t.Run("Transaction should not grow uncontrollably in size", func(t *testing.T) {
+		contractLen := 1000
+		contractCode := make([]byte, contractLen)
```
psiemens, line 33: Just to double check -- does the zeroed byte array translate to a non-empty hex string? e.g. `"000000..."` instead of just `""` or `"0"`

turbolent:
I don't think so, this is purely a Go SDK feature and the API of the template function hasn't changed.

psiemens (approved):
Looks good! This doesn't add new functionality and there are no specific docs to change. We should just make sure that this change is covered in the release notes/changelog.

One last thing, I think the `bytesToCadenceArray` function can be deleted now

turbolent:
The optimizations look good, it's a bit unfortunate that we're loosing the type-safety though.

As far as I remember, we talked about maybe having a "Bytes"/"Data" type, for which we could have an efficient Cadence-JSON encoding too. I'll create an issue for it

templates/accounts_test.go
janezpodhostnik: yes: https://play.golang.org/p/-BlkWurnsrv

I could make them random, but I am not comfortable with random elements in tests (not that it would matter here).

psiemens: @janezpodhostnik Nope that's okay! I just wasn't sure what the behaviour was and wanted to check.

turbolent:
Opened onflow/cadence#452 to improve this

PR #114, opened 2020-11-21
Allow passing options to KMS client constructor.

10thfloor:
## Description
Allow for configuring google KMS client with options object.

### DIFF
--- a/crypto/cloudkms/cloudkms.go
+++ b/crypto/cloudkms/cloudkms.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 
 	kms "cloud.google.com/go/kms/apiv1"
+	"google.golang.org/api/option"
 	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
 
 	"github.com/onflow/flow-go-sdk/crypto"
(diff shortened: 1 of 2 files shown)


crypto/cloudkms/cloudkms.go
```diff
@@ -89,8 +90,8 @@ type Client struct {
 }
 
 // NewClient creates a new KMS client.
-func NewClient(ctx context.Context) (*Client, error) {
-	client, err := kms.NewKeyManagementClient(ctx)
+func NewClient(ctx context.Context, opts option.ClientOption) (*Client, error) {
+	client, err := kms.NewKeyManagementClient(ctx, opts)
```
psiemens, line 94:
```suggestion
func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
	client, err := kms.NewKeyManagementClient(ctx, opts...)
```

The `...` operator is needed to turn this into a variadic argument that can be passed like this:

```go
c1 := NewClient(ctx, opt1)
c2 := NewClient(ctx, opt1, opt2)
```

psiemens (dismissed):
Nice catch! Have you been making use of the KMS package?

psiemens (approved):

Issue #115, opened 2020-11-21
Add ability to decode RLP encoded transactions


Kay-Zee:
**Is your feature request related to a problem? Please describe.**
Some easy method of passing transaction data to and from software using the SDK is required to allow things like multi-sig transactions to be done more easily across machines/services.

Currently, there is a way to RLP encode the transactions, but no easy way to decode it in the SDK

**Describe the solution you'd like**
Add RLP decoding, which would just reverse the encoding

**Describe alternatives you've considered**
Only thing to consider is how the canonical forms are handled. Whether they should stay as embedded structs, or be pulled out as private structs

**Additional context**
Blocker for easy multi sig transactions using go sdk

PR #116, opened 2020-11-21
Add transaction decoding

Kay-Zee:
Closes: #115

## Description

Adds RLP decode functionality for transactions

Two main considerations here:
- Is this how we want to handle the canonical forms? I've pulled them out as private structs, but not sure if this is the desired implementation
- Why does the transaction signature have address, but not pass it along? This causes all of my envelope decode tests to fail

______

For contributor use:

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -410,8 +410,7 @@ func DecodeTransactionPayloadMessage(envelopeMessage []byte) (*Transaction, erro
 	for i, auth := range temp.Authorizers {
 		authorizers[i] = BytesToAddress(auth)
 	}
-	t := new(Transaction)
-	*t = Transaction{
+	t := &Transaction{
 		Script:           temp.Script,
 		Arguments:        temp.Arguments,
 		ReferenceBlockID: BytesToID(temp.ReferenceBlockID),
@@ -437,7 +436,6 @@ func DecodeTransactionPayloadMessage(envelopeMessage []byte) (*Transaction, erro
 //
 // This message is only signed by the payer account.
 func DecodeTransactionEnvelopeMessage(envelopeMessage []byte) (*Transaction, error) {
-	t := new(Transaction)
 	temp := envelopeCanonicalForm{}
 	mustRLPDecode(envelopeMessage, &temp)
 	authorizers := make([]Address, len(temp.Payload.Authorizers))
@@ -449,7 +447,7 @@ func DecodeTransactionEnvelopeMessage(envelopeMessage []byte) (*Transaction, err
 	for i, sig := range temp.PayloadSignatures {
 		payloadSignatures[i] = transactionSignatureFromCanonicalForm(sig)
 	}
-	*t = Transaction{
+	t := &Transaction{
 		Script:           temp.Payload.Script,
 		Arguments:        temp.Payload.Arguments,
 		ReferenceBlockID: BytesToID(temp.Payload.ReferenceBlockID),
@@ -479,7 +477,6 @@ func DecodeTransactionEnvelopeMessage(envelopeMessage []byte) (*Transaction, err
 
 // DecodeTransaction returns the signable message for the transaction.
 func DecodeTransaction(transactionMessage []byte) (*Transaction, error) {
-	t := new(Transaction)
 	temp := transactionCanonicalForm{}
 	mustRLPDecode(transactionMessage, &temp)
 	authorizers := make([]Address, len(temp.Payload.Authorizers))
@@ -487,7 +484,7 @@ func DecodeTransaction(transactionMessage []byte) (*Transaction, error) {
 		fmt.Println(auth)
 		authorizers[i] = BytesToAddress(auth)
 	}
-	*t = Transaction{
+	t := &Transaction{
 		Script:           temp.Payload.Script,
 		Arguments:        temp.Payload.Arguments,
 		ReferenceBlockID: BytesToID(temp.Payload.ReferenceBlockID),
--- a/transaction.go
+++ b/transaction.go
@@ -557,12 +557,11 @@ func (s TransactionSignature) canonicalForm() transactionSignatureCanonicalForm 
 	}
 }
 
-func transactionSignatureFromCanonicalForm(v interface{}) TransactionSignature {
-	temp := v.(transactionSignatureCanonicalForm)
+func transactionSignatureFromCanonicalForm(v transactionSignatureCanonicalForm) TransactionSignature {
 	return TransactionSignature{
-		SignerIndex: int(temp.SignerIndex),
-		KeyIndex:    int(temp.KeyIndex),
-		Signature:   temp.Signature,
+		SignerIndex: int(v.SignerIndex),
+		KeyIndex:    int(v.KeyIndex),
+		Signature:   v.Signature,
 	}
 }
 
--- a/transaction.go
+++ b/transaction.go
@@ -328,23 +328,25 @@ func (t *Transaction) PayloadMessage() []byte {
 	return mustRLPEncode(&temp)
 }
 
-func (t *Transaction) payloadCanonicalForm() interface{} {
+type payloadCanonicalForm struct {
+	Script                    []byte
+	Arguments                 [][]byte
+	ReferenceBlockID          []byte
+	GasLimit                  uint64
+	ProposalKeyAddress        []byte
+	ProposalKeyIndex          uint64
+	ProposalKeySequenceNumber uint64
+	Payer                     []byte
+	Authorizers               [][]byte
+}
+
+func (t *Transaction) payloadCanonicalForm() payloadCanonicalForm {
 	authorizers := make([][]byte, len(t.Authorizers))
 	for i, auth := range t.Authorizers {
 		authorizers[i] = auth.Bytes()
 	}
 
-	return struct {
-		Script                    []byte
-		Arguments                 [][]byte
-		ReferenceBlockID          []byte
-		GasLimit                  uint64
-		ProposalKeyAddress        []byte
-		ProposalKeyIndex          uint64
-		ProposalKeySequenceNumber uint64
-		Payer                     []byte
-		Authorizers               [][]byte
-	}{
+	return payloadCanonicalForm{
 		Script:                    t.Script,
 		Arguments:                 t.Arguments,
 		ReferenceBlockID:          t.ReferenceBlockID[:],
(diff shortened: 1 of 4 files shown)


psiemens (commented):
> Is this how we want to handle the canonical forms? I've pulled them out as private structs, but not sure if this is the desired implementation

Yep, this sounds good to me. It was possible to define them as anonymous structs when the encoding was only one way, but it makes sense to pull them out now.

> Why does the transaction signature have address, but not pass it along? This causes all of my envelope decode tests to fail

The address is stripped from the canonical form and replaced with `signerIndex`, which is the index of the signer address in this list (returned by `tx.signerList()`):

```go
// signerList returns a list of unique accounts required to sign this transaction.
//
// The list is returned in the following order:
// 1. PROPOSER
// 2. PAYER
// 2. AUTHORIZERS (in insertion order)
```

This just reduces the size of the resulting RLP buffer. Sorry it's not directly documented.

I think the easiest fix is to construct the base transaction and then do something like this:

```go
signers := tx.signerList()

for _, sig := sigs {
  address := signers[sig.signerIndex]
}
```

I'd prefer to stay consistent with the other `Encode`/`Decode` functions in this package and do something like this:

```
func (tx *Transaction) Encode() []byte { ... }

func DecodeTransaction(b []byte) (*Transaction, error) { ... }
```

In fact, I think it might be possible to simplify the API so that only `DecodeTransaction` is needed. We'd need a top-level canonical form like this:

```go
type transactionCanonicalForm struct {
		Payload                     payloadCanonicalForm
		PayloadSignatures   []transactionSignatureCanonicalForm
		EnvelopeSignatures []transactionSignatureCanonicalForm
}
```

I could be wrong, but I think due to the nature of RLP, all three encoded versions (payload, envelope, full tx) will fit into this struct.

Kay-Zee:
> I think due to the nature of RLP, all three encoded versions (payload, envelope, full tx) will fit into this struct.

When I tried to rlpDecode into this struct, bytes generated by `tx.EnvelopeMessage` i end up getting:
```
	panic: rlp: too few elements for flow.transactionCanonicalForm
```
So I've kept the two other functions, `DecodeTransactionPayloadMessage` `DecodeTransactionEnvelopeMessage` I was debating a little bit on the naming, but I figured this was the most clear.

Your suggestion of:
```
signers := tx.signerList()

for _, sig := sigs {
  address := signers[sig.signerIndex]
}
```
worked great! so went with that.

Lastly, a thing to note is that there are two body fields are are possible to be empty, which are the script and argument fields. RLP decodes these as empty byte slices, but when we initialize them, we initialize them to nil. I've had to manually turn empty slices back to `nil` for these two fields.

I didn't bother checking the other fields, since I don't believe they can be empty and for this to still be a valid transaction, but let me know if you feel other wise.

transaction.go
```diff
@@ -307,23 +330,13 @@ func (t *Transaction) PayloadMessage() []byte {
...
 }
 
-func (t *Transaction) payloadCanonicalForm() interface{} {
+func (t *Transaction) payloadCanonicalForm() payloadCanonicalForm {
 	authorizers := make([][]byte, len(t.Authorizers))
 	for i, auth := range t.Authorizers {
 		authorizers[i] = auth.Bytes()
 	}
 
-	return struct {
```
turbolent, line 316: 👌

transaction.go
```diff
@@ -344,11 +357,8 @@ func (t *Transaction) EnvelopeMessage() []byte {
 	return mustRLPEncode(&temp)
 }
 
-func (t *Transaction) envelopeCanonicalForm() interface{} {
-	return struct {
```
turbolent, line 348: 👌

transaction.go
```diff
@@ -369,6 +379,134 @@ func (t *Transaction) Encode() []byte {
...
+		ReferenceBlockID: BytesToID(temp.ReferenceBlockID),
+		GasLimit:         temp.GasLimit,
+		ProposalKey: ProposalKey{
+			Address:        BytesToAddress(temp.ProposalKeyAddress),
+			KeyIndex:       int(temp.ProposalKeyIndex),
+			SequenceNumber: temp.ProposalKeySequenceNumber,
+		},
+		Payer:       BytesToAddress(temp.Payer),
+		Authorizers: authorizers,
+	}
```
turbolent, line 405: Maybe just the shorter:

```suggestion
	t := &Transaction{
		Script:           temp.Script,
		Arguments:        temp.Arguments,
		ReferenceBlockID: BytesToID(temp.ReferenceBlockID),
		GasLimit:         temp.GasLimit,
		ProposalKey: ProposalKey{
			Address:        BytesToAddress(temp.ProposalKeyAddress),
			KeyIndex:       int(temp.ProposalKeyIndex),
			SequenceNumber: temp.ProposalKeySequenceNumber,
		},
		Payer:       BytesToAddress(temp.Payer),
		Authorizers: authorizers,
	}
```

Or maybe I'm missing something

transaction.go
```diff
@@ -369,6 +379,134 @@ func (t *Transaction) Encode() []byte {
...
+		GasLimit:         temp.Payload.GasLimit,
+		ProposalKey: ProposalKey{
+			Address:        BytesToAddress(temp.Payload.ProposalKeyAddress),
+			KeyIndex:       int(temp.Payload.ProposalKeyIndex),
+			SequenceNumber: temp.Payload.ProposalKeySequenceNumber,
+		},
+		Payer:             BytesToAddress(temp.Payload.Payer),
+		Authorizers:       authorizers,
+		PayloadSignatures: payloadSignatures,
+	}
```
turbolent, line 444:
```suggestion
	temp := envelopeCanonicalForm{}
	mustRLPDecode(envelopeMessage, &temp)
	authorizers := make([]Address, len(temp.Payload.Authorizers))
	for i, auth := range temp.Payload.Authorizers {
		fmt.Println(auth)
		authorizers[i] = BytesToAddress(auth)
	}
	payloadSignatures := make([]TransactionSignature, len(temp.PayloadSignatures))
	for i, sig := range temp.PayloadSignatures {
		payloadSignatures[i] = transactionSignatureFromCanonicalForm(sig)
	}
	t := &Transaction{
		Script:           temp.Payload.Script,
		Arguments:        temp.Payload.Arguments,
		ReferenceBlockID: BytesToID(temp.Payload.ReferenceBlockID),
		GasLimit:         temp.Payload.GasLimit,
		ProposalKey: ProposalKey{
			Address:        BytesToAddress(temp.Payload.ProposalKeyAddress),
			KeyIndex:       int(temp.Payload.ProposalKeyIndex),
			SequenceNumber: temp.Payload.ProposalKeySequenceNumber,
		},
		Payer:             BytesToAddress(temp.Payload.Payer),
		Authorizers:       authorizers,
		PayloadSignatures: payloadSignatures,
	}
```

transaction.go
```diff
@@ -369,6 +379,134 @@ func (t *Transaction) Encode() []byte {
...
+		ReferenceBlockID: BytesToID(temp.Payload.ReferenceBlockID),
+		GasLimit:         temp.Payload.GasLimit,
+		ProposalKey: ProposalKey{
+			Address:        BytesToAddress(temp.Payload.ProposalKeyAddress),
+			KeyIndex:       int(temp.Payload.ProposalKeyIndex),
+			SequenceNumber: temp.Payload.ProposalKeySequenceNumber,
+		},
+		Payer:       BytesToAddress(temp.Payload.Payer),
+		Authorizers: authorizers,
+	}
```
turbolent, line 481:
```suggestion
	temp := transactionCanonicalForm{}
	mustRLPDecode(transactionMessage, &temp)
	authorizers := make([]Address, len(temp.Payload.Authorizers))
	for i, auth := range temp.Payload.Authorizers {
		fmt.Println(auth)
		authorizers[i] = BytesToAddress(auth)
	}
	t := &Transaction{
		Script:           temp.Payload.Script,
		Arguments:        temp.Payload.Arguments,
		ReferenceBlockID: BytesToID(temp.Payload.ReferenceBlockID),
		GasLimit:         temp.Payload.GasLimit,
		ProposalKey: ProposalKey{
			Address:        BytesToAddress(temp.Payload.ProposalKeyAddress),
			KeyIndex:       int(temp.Payload.ProposalKeyIndex),
			SequenceNumber: temp.Payload.ProposalKeySequenceNumber,
		},
		Payer:       BytesToAddress(temp.Payload.Payer),
		Authorizers: authorizers,
	}
```

psiemens:
> When I tried to rlpDecode into this struct, bytes generated by tx.EnvelopeMessage i end up getting:

	panic: rlp: too few elements for flow.transactionCanonicalForm
	
What do you think about manually decoding the RLP byte array? I'm still pretty into simplifying the API to have a single `DecodeTransaction` function.

However it's a lot more code and would look something like this:

```go

func DecodeTransaction(transactionMessage []byte) (*Transaction, error) {
  s := rlp.NewStream(bytes.NewReader(transactionMessage), 0)
    
  script, err := s.Bytes()
  if err != nil {
    return nil, err
  }
  
  
  argCount, err := s.List()
  if err != nil {
    return nil, err
  }
  
  args := make([][]byte, argCount)
  
  for i := 0; i < argCount; i++ {
    arg, err := s.Bytes()
    if err != nil {
      return nil, err 
    }
    
    args = append(args, arg)
  }
  
  err = s.ListEnd()
  if err != nil {
    return nil, err 
  }

  // ...
  
  return &Transaction{
    Script: script,
    Arguments: arguments,
    // ...
  }, nil
}
```

transaction.go
```diff
@@ -369,6 +379,131 @@ func (t *Transaction) Encode() []byte {
 	return mustRLPEncode(&temp)
 }
 
+// DecodeTransactionPayloadMessage returns the signable message for the transaction payload.
```
psiemens, line 382: The doc comments for the decode functions are wrong I think 

Maybe something like this?

> DecodeTransactionPayloadMessage constructs a transaction from an RLP-encoded payload message.

Kay-Zee:
While I agree with the concept of consolidating it down to one function, I wasn't sure if decoding each individual field manually was the best way to go/that maintainable. But thanks to your suggestion of using `rlp.Stream`, I think I've found a way around the previous blocker with the rlp number of fields error, and have consolidated all the decode functions into a `DecodeTransaction` function, that takes strong inspiration from your suggestion.

The main blocker with this implementation was how different the `payloadCanonicalForm` is in structure compared to the other two canonical forms. I've simply coded in a special case for it.

transaction.go
```diff
@@ -390,6 +403,133 @@ func (t *Transaction) Encode() []byte {
...
+		return nil, err
+	}
+
+	// Need to look at the type of the first element to determine if how we're going to be decoding
+	kind, _, err = s.Kind()
+	if err != nil {
+		return nil, err
+	}
+	// If first kind is not list, safe to assume this is actually just encoded payload, and decrypt as such
+	if kind != rlp.List {
```
psiemens, line 485: Nice! Really like this solution 👍

transaction.go
```diff
@@ -390,6 +403,133 @@ func (t *Transaction) Encode() []byte {
...
+	if len(t.Arguments) == 0 {
+		t.Arguments = nil
+	}
+	if len(t.Script) == 0 {
+		t.Script = nil
+	}
+	return t, nil
+}
+
+func decodeTransaction(transactionMessage []byte) (*transactionCanonicalForm, error) {
```
psiemens, line 460: Good idea to split this into a helper

psiemens (approved):
Great solution! Definitely feels cleaner to do the conditional decoding rather than each field manually. LGTM 🚢

Issue #113, opened 2020-11-18
crypto: use flow-go/crypto instead of internal


tarakby:
### Context

The SDK used its own copy of the crypto code because it got open-sourced while flow-go was still private. That shouldn't be the case anymore as flow-go became public.

### Definition of done

Delete flow-go-sdk/crytpo/internal and use onflow/flow-go/crypto instead.

bjartek:
It looks like the crypto/internal package is done now, why was this reopened?

tarakby:
It's a mistake, thanks for pointing out!

PR #118, opened 2020-11-24
Crypto: delete internal crypto 

tarakby:
Closes: https://github.com/onflow/flow-go-sdk/issues/113

## Description

- deletes the crypto internal folder
- uses `flow-go/crypto` without breaking changes to the SDK api. 

I tried to not break any external api in this PR, but a better refactoring and cleaning could be done if I'm allowed to introduce breaking changes and remove redundant types from the sdk (for instance remove `flow-go-sdk/crypto.PrivateKey` and use `flow-go/crypto.PrivateKey` instead, and a few other examples). Let me know if such changes are possible. cc @psiemens @turbolent


crypto/crypto.go
```diff
@@ -25,33 +25,29 @@ import (
...
 
-	"github.com/onflow/flow-go-sdk/crypto/internal/crypto"
-	"github.com/onflow/flow-go-sdk/crypto/internal/crypto/hash"
+	"github.com/onflow/flow-go/crypto"
+	"github.com/onflow/flow-go/crypto/hash"
 )
 
 // SignatureAlgorithm is an identifier for a signature algorithm (and parameters if applicable).
-type SignatureAlgorithm int
+type SignatureAlgorithm crypto.SigningAlgorithm
```
tarakby, line 33: This kind of duplicated types could be avoided if we don't mind deleting `flow-go-sdk/crypto.SignatureAlgorithm`. That's a breaking change though.

crypto/crypto.go
```diff
@@ -62,32 +58,29 @@ func StringToSignatureAlgorithm(s string) SignatureAlgorithm {
...
-	UnknownHashAlgorithm HashAlgorithm = iota
-	SHA2_256
-	SHA2_384
-	SHA3_256
-	SHA3_384
+	UnknownHashAlgorithm HashAlgorithm = HashAlgorithm(hash.UnknownHashingAlgorithm)
+	SHA2_256                           = HashAlgorithm(hash.SHA2_256)
+	SHA2_384                           = HashAlgorithm(hash.SHA2_384)
+	SHA3_256                           = HashAlgorithm(hash.SHA3_256)
+	SHA3_384                           = HashAlgorithm(hash.SHA3_384)
```
tarakby, line 68: I would like to remove the `384` versions as they are not used in Flow, but that's also considered a breaking change.

crypto/crypto.go
```diff
@@ -111,52 +104,27 @@ func CompatibleAlgorithms(sigAlgo SignatureAlgorithm, hashAlgo HashAlgorithm) bo
...
-func (sk PrivateKey) Sign(message []byte, hasher Hasher) ([]byte, error) {
-	return sk.privateKey.Sign(message, hasher)
+	crypto.PrivateKey
 }
 
 // Algorithm returns the signature algorithm for this private key.
 func (sk PrivateKey) Algorithm() SignatureAlgorithm {
-	return SignatureAlgorithm(sk.privateKey.Algorithm())
+	return SignatureAlgorithm(sk.PrivateKey.Algorithm())
 }
```
tarakby, line 113: The embedding avoids redefining the same functions, it doesn't work here because of the output type.

crypto/crypto.go
turbolent: I think you can say `type SignatureAlgorithm = crypto.SigningAlgorithm` here

crypto/crypto.go
```diff
@@ -197,12 +165,15 @@ func NewNaiveSigner(privateKey PrivateKey, hashAlgo HashAlgorithm) NaiveSigner {
...
+// targeted security bits of the cryptographic algorithms
+const securityBits = 128
+
 // MinSeedLength is the generic minimum seed length required to guarantee sufficient
 // entropy when generating keys.
 //
 // This minimum is used when the seed source is not necessarily a CSPRG and the seed
 // should be expanded before being passed to the key generation process.
-const MinSeedLength = crypto.MinSeedLen
+const MinSeedLength = 2 * (securityBits / 8)
```
turbolent, line 176: Maybe this should be a constant in flow-go/crypto?

turbolent (approved):
Nice!

crypto/crypto.go
tarakby: The crypto lib defines other types of minimum seed lengths because it makes other assumptions about the seed quality.
This constant is here to address another point when the SDK generates keys from passphrases.

crypto/crypto.go
tarakby: It's not possible because of [this function](https://github.com/onflow/flow-go-sdk/blob/tarak/use-flow-go-crypto/crypto/crypto.go#L44) is not allowed to declare function for a receiver from another package. 
The function could be totally deleted if we don't mind removing  `flow-go-sdk/crypto.SignatureAlgorithm`.

tarakby:
Hey @psiemens, do you have a preference about making breaking changes to the SDK crypto APIs ? If you think breaking changes are fine, I'll make some cleaner changes to this PR. If breaking changes are a problem, we can merge the PR in its current state. Thanks!

crypto/crypto.go
```diff
@@ -25,33 +25,24 @@ import (
...
 )
 
 // SignatureAlgorithm is an identifier for a signature algorithm (and parameters if applicable).
-type SignatureAlgorithm int
+type SignatureAlgorithm = crypto.SigningAlgorithm
 
 const (
-	UnknownSignatureAlgorithm SignatureAlgorithm = iota
-	// BLS_BLS12381 is BLS on BLS 12-381 curve
-	BLS_BLS12381
```
psiemens, line 38: 👍

flow.go
```diff
@@ -87,10 +87,10 @@ func (id ChainID) String() string {
 // entityHasher is a thread-safe hasher used to hash Flow entities.
 type entityHasher struct {
 	mut    sync.Mutex
-	hasher crypto.Hasher
+	hasher hash.Hasher
```
psiemens, line 90: Could these all stay as `crypto.Hash` and `crypto.Hasher`? 

I'd prefer to have the rest of the SDK files depend only on `flow-go-sdk/crypto`

crypto/crypto_test.go
```diff
@@ -36,7 +37,7 @@ func TestGeneratePrivateKey(t *testing.T) {
 
 	// key algorithms not currently supported by the SDK
 	unsupportedAlgos := []crypto.SignatureAlgorithm{
-		crypto.BLS_BLS12381,
+		fgcrypto.BLSBLS12381,
```
psiemens, line 40: 👍

psiemens (commented):
Looks good! Just added one point in a comment.

I'd prefer to avoid breaking changes for now, so the way you have it is great 👌

tarakby:
Is there a way other than `go mod tidy` to get a consistent `go.sum` ?  
The CI `make check-tidy` is showing a `go.sum` difference but `go mod tidy` doesn't seem to fix it 🤷‍♂️

turbolent:
@tarakby you probably just want to commit the changes in to go.sum file. If that's the output of go mod tidy, then that's what should be in the repo

psiemens (approved):
This looks good! I think you just need to do a rebase and re-run `go mod tidy` to clear up the conflict. I can do that if you'd like 😄

tarakby:
Yes please @psiemens, feel free to merge it. You can try with the current PR or [this one](https://github.com/onflow/flow-go-sdk/pull/138), it has the same changes but a cleaner history, my merging and rebasing doesn't seem to work 😅

Issue #125, opened 2020-12-10
Client: Add the ability to pass in gRPC call options


yndc:
**Is your feature request related to a problem? Please describe.**
I can't find a way to add in gRPC call options for `client.Client` methods right now, this made retrieving some data from the blockchain impossible for huge blocks / transactions.

For example, calling `GetTransactionResult` for the transaction with the hash `b25c5a82b4dc9d6dfd2f7eb2016d272ae4b8f03d65a1803c437b046547f3893f` in the testnet would cause an error: `client: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (7714488 vs. 4194304)`

The solution is to increase the gRPC inbound message maximum limit with `grpc.MaxCallRecvMsgSize`, but there's no way to do this at the moment.

**Describe the solution you'd like**
Provide a way to pass in gRPC call options for several `client.Client` methods where message size could be an issue (`GetTransactionResult` in my case).

Something like `func (c *Client) GetTransactionResult(ctx context.Context, txID flow.Identifier, opts ...grpc.CallOption)` so that it won't be a breaking change.

PR #126, opened 2020-12-11
Add call options parameter to all Access API gRPC client functions

psiemens:
Fixes: #125

## Description

This PR adds the `...grpc.CallOption` parameter to all Access API gRPC client functions on `client.Client`, allowing the caller to configure the functionality of the underlying gRPC client.

cc @yndc
______

For contributor use:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -83,18 +83,23 @@ func (c *Client) Close() error {
 }
 
 // Ping is used to check if the access node is alive and healthy.
-func (c *Client) Ping(ctx context.Context) error {
-	_, err := c.rpcClient.Ping(ctx, &access.PingRequest{})
+func (c *Client) Ping(ctx context.Context, opts ...grpc.CallOption) error {
+	_, err := c.rpcClient.Ping(ctx, &access.PingRequest{}, opts...)
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


turbolent (approved):
Great idea! 👍

awfm9:
I looked at the PR without any background, but it did strike me as having a lot of redundant parameters in each function.

Did you consider doing something like this?

```go
client.WithOptions(opts...).Ping(ctx)
client.WithOptions(opts...).GetLatestBlockHeader(ctx, true)
```

In fact, looking at it now, I don't see a reason not to extend it to the context itself:

```go
client.Ping()
client.WithContext(ctx).GetLatestBlockHeader(true)
client.WithOptions(opts...).GetTransactionResult(txID)
client.WithContext(ctx).WithOptions(opts...).GetEventsForHighRange(query)
```

@psiemens what do you think?

psiemens:
@awfm9 I thought about implementing it using something like `WithOptions`, but after looking at the [grpc godoc](https://godoc.org/google.golang.org/grpc#CallOption), it looks like some of the options are only meant to be used once (e.g. `grpc.Header(&h)` mutates a value at a pointer). I didn't want to stray too far from the underlying gRPC implementation in case there were unintended side effects (now or in the future).

PR #134, opened 2021-01-15
Update SECURITY.md

laynelafrance:
@jkan2 do you know where the doc we reference there lives for me to add the link?


jkan2:
@laynelafrance i don't think we have one yet

psiemens:
@laynelafrance I think this line is just saying that any future contributions will be recognized "here", as in inside this `SECURITY.md` document.

PR #135, opened 2021-01-18
Storage fees and account minimum balance example

janezpodhostnik:
Closes: https://github.com/dapperlabs/flow-internal/issues/1369

## Description

Added an example for storage:
- transaction tries to store a large resource
- demonstrates that acct.storageUsed will change after adding resource
- demonstrates that acct.storageUsed can be compared to acct.stroageCapacity
- demonstrates the the error for going over limit is predictable
______

For contributor use:

### DIFF
--- a/examples/deploy_contract/main.go
+++ b/examples/deploy_contract/main.go
@@ -1,7 +1,7 @@
 /*
  * Flow Go SDK
  *
- * Copyright 2019-2020 Dapper Labs, Inc.
+ * Copyright 2019-2021 Dapper Labs, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -93,18 +93,20 @@ func DeployContractDemo() {
 		[]templates.Contract{{
 			Name:   "GreatToken",
 			Source: nftCode,
-		}}, myAddress)
+		}}, serviceAcctAddr)
 
-	deployContractTx.SetProposalKey(myAddress, myAcctKey.Index, myAcctKey.SequenceNumber)
+	deployContractTx.SetProposalKey(
+		serviceAcctAddr,
+		serviceAcctKey.Index,
+		serviceAcctKey.SequenceNumber,
+	)
 	// we can set the same reference block id. We shouldn't be to far away from it
 	deployContractTx.SetReferenceBlockID(referenceBlockID)
-	deployContractTx.SetPayer(myAddress)
+	deployContractTx.SetPayer(serviceAcctAddr)
 
-	err = deployContractTx.SignEnvelope(
-		myAddress,
-		myAcctKey.Index,
-		crypto.NewInMemorySigner(myPrivateKey, myAcctKey.HashAlgo),
-	)
+
+	err = deployContractTx.SignEnvelope(serviceAcctAddr, serviceAcctKey.Index, serviceSigner)
+	examples.Handle(err)
 	examples.Handle(err)
 
 	err = flowClient.SendTransaction(ctx, *deployContractTx)
@@ -114,7 +116,7 @@ func DeployContractDemo() {
 	examples.Handle(deployContractTxResp.Error)
 
 	// Successful Tx, increment sequence number
-	myAcctKey.SequenceNumber++
+	serviceAcctKey.SequenceNumber++
 
 	var nftAddress flow.Address
 
--- a/examples/storage_usage/main.go
+++ b/examples/storage_usage/main.go
@@ -1,0 +1,164 @@
+/*
+ * Flow Go SDK
+ *
+ * Copyright 2019-2021 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


examples/storage_usage/main.go
```diff
@@ -0,0 +1,171 @@
...
+		keySigner)
+
+	result := examples.WaitForSeal(ctx, flowClient, txId)
+
+	if result.Error == nil {
+		fmt.Println("Storage limits are off")
+		return
+	}
+
+	if !strings.HasPrefix(result.Error.Error(), "execution error code 11") {
```
turbolent, line 95: "error code 11" seems a bit like a magic number. Would it be possible to improve this check so it gets tries to get the code of the error, and if it is an error with a code, compares the code with a constant (that has a descriptive name)?

examples/storage_usage/main.go
```diff
@@ -0,0 +1,171 @@
...
+				
+				// create resource and save it on the account 
+				let bigResource <- StorageDemo.createStorageTestResource("%s")
+				acct.save(<-bigResource, to: /storage/StorageDemo)
+
+				let storageUsedAfter = acct.storageUsed
+
+				if (storageUsed == storageUsedAfter){
+					panic("storage used will change")
+				}
```
turbolent, line 132:
```suggestion
				if storageUsed == storageUsedAfter {
					panic("storage used will change")
				}
```

examples/storage_usage/main.go
```diff
@@ -0,0 +1,171 @@
...
+				
+				if (storageUsedAfter >  acct.storageCapacity) {
```
turbolent, line 134:
```suggestion
				if storageUsedAfter > acct.storageCapacity {
```

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/btcsuite/btcd v0.0.0-20171128150713-2e60448ffcc6
 	github.com/ethereum/go-ethereum v1.9.9
 	github.com/golang/protobuf v1.4.2
-	github.com/onflow/cadence v0.10.2
+	github.com/onflow/cadence v0.10.5
```
turbolent, line 10: Should this branch target an older version of the SDK? `master` is already at 0.12: https://github.com/onflow/flow-go-sdk/blob/master/go.mod#L10

turbolent (commented):
Nice work on the examples, they'll be useful to understand the feature

examples/deploy_contract/main.go
```diff
@@ -93,18 +93,20 @@ func DeployContractDemo() {
...
+	)
 	// we can set the same reference block id. We shouldn't be to far away from it
 	deployContractTx.SetReferenceBlockID(referenceBlockID)
-	deployContractTx.SetPayer(myAddress)
+	deployContractTx.SetPayer(serviceAcctAddr)
 
-	err = deployContractTx.SignEnvelope(
-		myAddress,
-		myAcctKey.Index,
-		crypto.NewInMemorySigner(myPrivateKey, myAcctKey.HashAlgo),
```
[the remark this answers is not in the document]
janezpodhostnik: Its a bit tricky actually! if minimum account balance is on, the newly created has 0.1 FLOW. If this account tries to create an account it needs 0.1 FLOW, but it cant use its own flow for that, so the second transaction fails. 

The solution is to add FLOW to the account, but there is no example that uses (or knows about) the FungibleToken address. I wasn't sure how to approach that.

I think the examples should have an `examples.addFLOWToAddress` method which calls a minting transaction, but I need to hardcode the FungibleToken and FlowToken addresses. Do you think that is ok?

examples/storage_usage/main.go
janezpodhostnik: So a function that looks for "code #" and returns the number if found. Should that be in the examples?

turbolent: Hmm, I meant on a type/field level, the function/statement shouldn't parse the error message/string. 

Where is code 11 defined / what does it mean?

go.mod
janezpodhostnik: Yes it should be. Changed to v0.12.3 for now. going to create v0.12.4 branch from v0.12.3  and point it to that in a moment.

examples/deploy_contract/main.go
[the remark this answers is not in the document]
turbolent: Hmm, I don't think it's simpler if we just assume that the account that attempts to deploy the contract has enough FLOW / prepared the target account, so that it has enough storage capacity.
We could just add a note/comment at the top of the example that points out this assumption, maybe something like:

```
// NOTE: This example assumes that the target account has enough storage capacity to store the source code and instantiated contract
``` 

I'm worried that the use of the service account, or minting and adding FLOW, will confuse users/developers, because both aren't things that they are able to do.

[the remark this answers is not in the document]
janezpodhostnik: Ok let me give you slightly more detail, cause I suspect we might not be 100% on the same page:

That example does (S -service account, A# - other accounts):
1. S creates A1
2. A1 creates A2
3. ... (this isn't relevant for now)

With storage fees this means:
1. S creates A1 and adds the minimum deposit (0.1 FLOW) to A1 
2. A1 creates A2 and adds the minimum deposit to A2, but this means A1 doesn't have the required minimum anymore! So this TX fails

This will always happen (regardless of how large the contract is that is being deployed), unless A1 has more than the minimum deposit.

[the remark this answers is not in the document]
turbolent: Sounds good! I just wanted to make sure we don't over complicate the simple example of deploying a contract

go.mod
turbolent: Perfect, thanks! Once this PR is merged, we should re-open it against master

turbolent (approved):
Nice!

PR #140, opened 2021-01-27
Add docs folder.

10thfloor:
Adding this folder for pulling markdown docs content directly into Gatsby site.

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -1,51 +1,11 @@
---- 
-title: Flow GO SDK 
+---
+title: Flow GO SDK
 description: Packages for Go developers to build applications that interact with the Flow network.
 ---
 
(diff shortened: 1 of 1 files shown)


psiemens (approved):

janezpodhostnik (approved):

turbolent:
The content in the docs directory is a copy of the README right? If so, this will probably get outdated. Maybe just remove move most of what is in the docs from the README and link to the docs instead

10thfloor:
> The content in the docs directory is a copy of the README right? If so, this will probably get outdated. Maybe just remove move most of what is in the docs from the README and link to the docs instead

Yes good idea. For now , we'll leave the README as it is – pending a quick discussion with repo maintainers about what we would like to see on the repo README... after which time we can replace it.

PR #141, opened 2021-02-01
Add templates.AddAccountContracts template

psiemens:
## Description

This PR adds the `template.AddAccountContracts` function, which constructs a transaction that adds or updates multiple contracts on an account.
______

For contributor use:

### DIFF
--- a/templates/accounts.go
+++ b/templates/accounts.go
@@ -20,6 +20,7 @@ package templates
 
 import (
 	"encoding/hex"
+
 	"github.com/onflow/cadence"
 	jsoncdc "github.com/onflow/cadence/encoding/json"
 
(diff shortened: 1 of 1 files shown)


templates/accounts.go
```diff
@@ -146,6 +147,43 @@ transaction(publicKey: String) {
 }
 `
 
+const addAccountContractsTemplate = `
```
turbolent, line 150: Maybe rename this to "addOrUpdate", to indicate that this template may also update an existing contract (which is not safe in all cases!).

Maybe alternatively remove the update branch and add another separate template for it

janezpodhostnik (approved):
I'm also worried about the update part. Updating a contract might produce unwanted results, so it should probably be a separate method...

Other than that looks good to me.

psiemens:
Closing this since it's no longer needed. Thanks for the review, though!

PR #175, opened 2021-03-21
Fix PEM decoding issue for some KMS keys

Kay-Zee:
Closes: N/A

## Description

Was running into a PEM decode issue when working with some KMS keys. Kept getting `cryptokms: failed to parse PEM public key: input has incorrect ECDSA_P256 key size`.

The cause of this is sometimes, when calling `.Bytes()` on the X and Y points after decoding the PEM key, the corresponding byte slices were not the full 32 bytes, which was expected by the next function (expecting ECDSA_P256 keys to have a raw key of length 64)

This adds left padding (since the bytes are returned big endian) up to the expected length for ECDSA_P256 only. Can add other known lengths as we start to use them more.

______

For contributor use:

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -247,46 +247,37 @@ func DecodePublicKeyHex(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
 	return DecodePublicKey(sigAlgo, b)
 }
 
-// It is possible in some cases where the resulting byte slice from the PEM
-// decoding of the X and Y points do not generate the right size byte slices
-// If we know there is an expected length for a specific signature algorithm,
-// we make sure to left pad it to the right length. This is a mapping of the
-// signature algorithms to the expected length of the byte slices representing
-// the X and Y points
-var expectedPointByteLength = map[SignatureAlgorithm]int{
-	crypto.ECDSAP256: 32,
-}
-
-// DecodePublicKeyHex decodes a PEM public key with the given signature algorithm.
+// DecodePublicKeyHex decodes a PEM ECDSA public key with the given curve.
 func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error) {
-	block, _ := pem.Decode([]byte(s))
+	block, rest := pem.Decode([]byte(s))
+	if len(rest) > 0 {
+		return PublicKey{}, fmt.Errorf("crypto: failed to parse PEM string, all not bytes in PEM key were decoded: %s", string(rest))
+	}
 
+	// TODO: Replace with function that is compatible with secp256k1
 	publicKey, err := x509.ParsePKIXPublicKey(block.Bytes)
 	if err != nil {
 		return PublicKey{}, fmt.Errorf("crypto: failed to parse PEM string: %w", err)
 	}
 
-	goPublicKey := publicKey.(*ecdsa.PublicKey)
+	goPublicKey, ok := publicKey.(*ecdsa.PublicKey)
+	if !ok {
+		return PublicKey{}, fmt.Errorf("only ECDSA public keys are supported")
+	}
 	xBytes := goPublicKey.X.Bytes()
 	yBytes := goPublicKey.Y.Bytes()
-	expectedLength := expectedPointByteLength[sigAlgo]
-	var rawPublicKey []byte
-	if expectedLength > 0 {
-		// If an expected length for the point byte slice sizes, make sure to
-		// pad up to the expected length
-		rawPublicKey = make([]byte, 0, 2*expectedLength)
-		rawPublicKey = appendWithLeftPad(rawPublicKey, xBytes, expectedLength)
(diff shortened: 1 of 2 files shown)


janezpodhostnik (approved):

turbolent:
@tarakby Can you please have a look here?

crypto/crypto.go
```diff
@@ -257,11 +267,31 @@ func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
 	}
 
 	goPublicKey := publicKey.(*ecdsa.PublicKey)
```
tarakby, line 269: I haven't noticed this before, we should test the type assertion here as other types of public keys may be returned. 
```
goPublicKey, ok := publicKey.(*ecdsa.PublicKey)
if !ok {
   return PublicKey{}, fmt.Error("only ECDSA public keys are supported")
}
```

crypto/crypto.go
```diff
@@ -247,6 +247,16 @@ func DecodePublicKeyHex(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
...
+// decoding of the X and Y points do not generate the right size byte slices
+// If we know there is an expected length for a specific signature algorithm,
+// we make sure to left pad it to the right length. This is a mapping of the
+// signature algorithms to the expected length of the byte slices representing
+// the X and Y points
+var expectedPointByteLength = map[SignatureAlgorithm]int{
+	crypto.ECDSAP256: 32,
+}
+
 // DecodePublicKeyHex decodes a PEM public key with the given signature algorithm.
```
tarakby, line 260:
```suggestion
// DecodePublicKeyHex decodes a PEM ECDSA public key with the given curve. 
```

crypto/crypto.go
```diff
@@ -247,6 +247,16 @@ func DecodePublicKeyHex(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
...
 func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error) {
 	block, _ := pem.Decode([]byte(s))
```
tarakby, line 262: I think we shouldn't have trailing data here.
```suggestion
	block, rest := pem.Decode([]byte(s))
	if len(rest)>0 {
	      return PublicKey{}, fmt.Errorf("crypto: failed to parse PEM string: %w", err)
	}
```

crypto/crypto.go
```diff
@@ -257,11 +267,31 @@ func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
 	}
 
 	goPublicKey := publicKey.(*ecdsa.PublicKey)
-
-	rawPublicKey := append(
-		goPublicKey.X.Bytes(),
-		goPublicKey.Y.Bytes()...,
-	)
+	xBytes := goPublicKey.X.Bytes()
```
tarakby, line 270: using `ParsePKIXPublicKey` only works in the case of curve P256. We will need to write a local `ParsePKIXPublicKey` function that supports secp256k1. I can help writing it.

crypto/crypto.go
```diff
@@ -257,11 +267,31 @@ func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
...
+	yBytes := goPublicKey.Y.Bytes()
+	expectedLength := expectedPointByteLength[sigAlgo]
```
tarakby, line 272: It's better to get that length this way:
```suggestion
	expectedLength := bitsToBytes(goPublicKey.Params().P.BitLen())
```
With `bitsToBytes` the function converting bit size into byte size:
```
func bitsToBytes(bits int) int {
	return (bits + 7) >> 3
}
```

crypto/crypto.go
```diff
@@ -257,11 +267,31 @@ func DecodePublicKeyPEM(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
...
+	var rawPublicKey []byte
+	if expectedLength > 0 {
```
tarakby, line 274: We can remove this check and pad the expected length in all cases.

crypto/crypto.go
```diff
@@ -247,6 +247,16 @@ func DecodePublicKeyHex(sigAlgo SignatureAlgorithm, s string) (PublicKey, error)
 	return DecodePublicKey(sigAlgo, b)
 }
 
+// It is possible in some cases where the resulting byte slice from the PEM
+// decoding of the X and Y points do not generate the right size byte slices
+// If we know there is an expected length for a specific signature algorithm,
+// we make sure to left pad it to the right length. This is a mapping of the
+// signature algorithms to the expected length of the byte slices representing
+// the X and Y points
+var expectedPointByteLength = map[SignatureAlgorithm]int{
```
tarakby, line 256: Eventually, we won't the map.

tarakby (commented):
I think there a few changes needed, some of them are related to the original function that existed before and I didn't notice. 
Happy to help make the changes.

PR #182, opened 2021-04-08
[Crypto] Simplify nested key structures

tarakby:
This PR doesn't address any ticket.

- simplifies the private and public key structures by using the `flow-go/crypto` structures.
- cleans up redundant functions.


turbolent (approved):
Nice cleanup

psiemens (approved):
Thanks for the links you sent about aliases in Go -- this change looks good to me.

Issue #61, opened 2020-06-18
Include domain tag when signing transactions


psiemens:
When signing a transaction, the message being signed should be prefixed with the transaction domain tag (TBD).

The tag should be prepended to the byte slice returned by the `PayloadMessage()` and `EnvelopeMessage()` functions on `flow.Transaction`.

## Definition of Done

- [ ] `PayloadMessage` produces a message that is prefixed with the transaction domain tag
- [ ] `EnvelopeMessage` produces a message that is prefixed with the transaction domain tag

cc @tarakby

janezpodhostnik:
Is this something every SDK will need? I assume so. Where can I read more about this? is this connected to a flow-go change?

studio1vc:
@psiemens is this planned for Sprint 34?

PR #184, opened 2021-04-13
Introduce transaction domain tag

janezpodhostnik:
Closes: #61 

## Description

Added the transaction domain tag before signing the transaction. The transaction domain tag can be changed, but it has to be changed by all signing parties (if the tx is encoded/decoded when passed from one signer to another signer).

______

For contributor use:


transaction.go
```diff
@@ -74,17 +74,24 @@ type Transaction struct {
 
 	// PayloadSignatures is a list of signatures generated by the proposer and authorizer roles.
 	//
-	// A payload signature is generated over the inner portion of the transaction (payload).
+	// A payload signature is generated over the inner portion of the transaction (TransactionDomainTag + npayload).
```
turbolent, line 77:
```suggestion
	// A payload signature is generated over the inner portion of the transaction (TransactionDomainTag + payload).
```

turbolent:
@tarakby can you please have a look here?

transaction.go
```diff
@@ -74,17 +74,24 @@ type Transaction struct {
...
 	//
 	// You can find more information about transaction signatures here: https://docs.onflow.org/concepts/transaction-signing/#anatomy-of-a-transaction
 	EnvelopeSignatures []TransactionSignature
+
+	// TransactionDomainTag is prepended to the transaction message before it is signed.
+	//
+	// The transaction domain tag is not encoded together with the transaction. The default transaction domain tag is `paddedDomainTag("FLOW-V0.0-transaction")`
+	//
+	// You can find more information about transaction domain tag here: https://docs.onflow.org/concepts/transaction-signing/#anatomy-of-a-transaction
+	TransactionDomainTag []byte
```
tarakby, line 94: `TransactionDomainTag` is a protocol-wide constant, rather than a transaction-specific field. I don't think it should be added to the transaction structure. 

The constant can be used directly in `SignEnvelope` and `SignPayload`.

transaction.go
```diff
@@ -304,7 +315,9 @@ func (t *Transaction) SignPayload(address Address, keyIndex int, signer crypto.S
 //
 // This function returns an error if the signature cannot be generated.
 func (t *Transaction) SignEnvelope(address Address, keyIndex int, signer crypto.Signer) error {
-	sig, err := signer.Sign(t.EnvelopeMessage())
+	message := t.EnvelopeMessage()
+	message = append(t.TransactionDomainTag, message...)
+	sig, err := signer.Sign(message)
```
tarakby, line 320: The comment doesn't belong to the code above:
Does the SDK also implement verifications for the payload and envelope signatures? If yes, then the tag needs to be prepended to the message in those functions as well.

transaction.go
janezpodhostnik: the `TransactionDomainTag` has an associated version: `paddedDomainTag("FLOW-V0.0-transaction")`. Is there no chance that in the future if this version changes, someone will want to use the new version before the sdk is updated (or without updating the sdk)?

tarakby: It's still not clear to me what the versioning will be used for (versions of the transaction model, versions of Flow forks, ..). For now the FMV only supports a unique static domain tag when verifying signatures, and not a dynamic one read on the transaction. If we were to enable multiple version support, we could have a specific field for for the version in the transaction structure instead of the entire tag, and that field will be checked by the FVM (and collection nodes) for transaction validity.  

Btw, I think it's a good idea to make the SDK re-use the [tag used by the FVM ](https://github.com/onflow/flow-go/blob/master/model/flow/constants.go#L43).

janezpodhostnik: > It's still not clear to me what the versioning will be used for (versions of the transaction model, versions of Flow forks, ..). For now the FMV only supports a unique static domain tag when verifying signatures, and not a dynamic one read on the transaction. If we were to enable multiple version support, we could have a specific field for for the version in the transaction structure instead of the entire tag, and that field will be checked by the FVM (and collection nodes) for transaction validity.

I understand. I will change it as you propose.

> Btw, I think it's a good idea to make the SDK re-use the tag used by the FVM .

`flow-go-sdk` does not depend on flow-go. In fact flow-go depends on the `flow-go-sdk`. So If anything, I could change `flow-go` to use the tag defined in the sdk, but that does not seem like a good idea.

tarakby (approved):
🏄‍♂️

client/convert/protobuf.go
```diff
@@ -444,7 +444,7 @@ func TransactionToMessage(t flow.Transaction) (*entities.Transaction, error) {
 
 func MessageToTransaction(m *entities.Transaction) (flow.Transaction, error) {
 	if m == nil {
-		return flow.Transaction{}, ErrEmptyMessage
+		return *flow.NewTransaction(), ErrEmptyMessage
```
psiemens, line 447: Why the change from struct literals to a constructor? Is that still needed now that the `TransactionDomainTag` field is removed?

transaction.go
psiemens: No, the Go SDK doesn't have any transaction signature verification functionality.

client/convert/protobuf.go
psiemens: If possible, I would change back just to keep this PR minimal.

psiemens (commented):
Nice!

transaction.go
tarakby: Thanks @Peter!
Do you know if there is a flow-go issue to mirror this SDK change in the verifications on the FVM side  ?

transaction.go
tarakby: Hm right, the only dependency I know in the opposite way  is `flow-go-sdk/crypto` depending on `flow-go/crypto`. 
I'll add the change on `flow-go` then.

transaction.go
janezpodhostnik: @tarakby this is the change in flow-go https://github.com/onflow/flow-go/pull/606

PR #211, opened 2021-09-01
Add non-zero default transaction gas limit

janezpodhostnik:
## Description

Due to https://github.com/onflow/flow-go/pull/1220 a gas limit of 0 is no longer a valid gas limit.

I changed the default gas limit (when you call `NewTransaction()` to 1000, which is enough for all of the examples.

Any comments on the default gas limit amount are very welcome. 

This relates to this emulator PR https://github.com/onflow/flow-emulator/pull/78.
______

For contributor use:

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -110,9 +110,14 @@ type transactionCanonicalForm struct {
 	EnvelopeSignatures []transactionSignatureCanonicalForm
 }
 
+// DefaultTransactionGasLimit should be high enough for small transactions
+const DefaultTransactionGasLimit = 1000
+
 // NewTransaction initializes and returns an empty transaction.
 func NewTransaction() *Transaction {
-	return &Transaction{}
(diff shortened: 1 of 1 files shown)


transaction.go
```diff
@@ -110,9 +110,14 @@ type transactionCanonicalForm struct {
 	EnvelopeSignatures []transactionSignatureCanonicalForm
 }
 
+// DefaultTransactionGasLimit should be high enough for small transactions
+const DefaultTransactionGasLimit = 1000
```
turbolent, line 114: Maybe this should match the behavior of the flow-go, which is to use the default limit? 9999 as far as I remember

janezpodhostnik: I didn't want to use the maximum as the default... but I guess right now there is no reason not to?

sideninja (approved):
LGTM thanks !

PR #214, opened 2021-09-21
Message signing and verification examples

janezpodhostnik:
## Description

Updated the existing examples with more comments and added another example for message signature verification, with the public keys already on an account.

______

For contributor use:

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,5 +1,5 @@
 .PHONY: all
-all: create-account add-account-key deploy-contract query-events storage-usage transaction-arguments single-party single-party-multisig multi-party multi-party-multisig user-signature
+all: create-account add-account-key deploy-contract query-events storage-usage transaction-arguments single-party single-party-multisig multi-party multi-party-multisig user-signature user-signature-from-account
 
 .PHONY: create-account
 create-account:
@@ -52,3 +52,7 @@ multi-party-multisig:
 .PHONY: user-signature
 user-signature:
 	go run ./user_signature/main.go
+
+.PHONY: user-signature-from-account
+user-signature-from-account:
+	go run ./user_signature_from_account/main.go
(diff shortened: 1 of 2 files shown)


examples/user_signature_from_account/main.go
```diff
@@ -0,0 +1,155 @@
...
+			if key.isRevoked {
+				continue
+			}
+			keyList.add(
+				PublicKey(
+					publicKey: key.publicKey.publicKey,
+					signatureAlgorithm: SignatureAlgorithm.ECDSA_P256
+				),
+				hashAlgorithm: key.hashAlgorithm,
+				weight: key.weight / 1000.0,
```
janezpodhostnik, line 64: @turbolent It seems a bit weird that if I want to use the public keys on an account for signature verification, I need to do a division by the threshold. Is there another way to do this?

examples/user_signature_from_account/main.go
```diff
@@ -0,0 +1,155 @@
...
+	let keyList = Crypto.KeyList()
+	
+	let account = getAccount(address)
+	let keys = account.keys
+	
+	var i = 0
+	while true {
+		if let key = keys.get(keyIndex: i) {
+			if key.isRevoked {
+				continue
```
janezpodhostnik, line 56: As far as I understand the revoked keys need to be skipped here. But if they are skipped the signature indexes will be off, right? How do I address that? 

Should I keep track of revoked keys and adjust the incoming `keyIndexes` accordingly? @turbolent

examples/Makefile
```diff
@@ -52,3 +52,7 @@ multi-party-multisig:
 .PHONY: user-signature
 user-signature:
 	go run ./user_signature/main.go
+
+.PHONY: user-signature-from-account
+user-signature-from-account:
```
sideninja, line 57: This new example is missing from README.md

examples/user_signature_from_account/main.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+			keyList.add(
+				PublicKey(
+					publicKey: key.publicKey.publicKey,
+					signatureAlgorithm: SignatureAlgorithm.ECDSA_P256
```
sideninja, line 61: Better to not hardcode
```suggestion
					signatureAlgorithm: key.publicKey.signatureAlgorithm
```

Issue #250, opened 2022-02-17
Signature in Block struct


HarleyAppleChoi:
**Is your feature request related to a problem? Please describe.**
In access API the [Block](https://github.com/onflow/flow-go-sdk/blob/573abf0bffa1334267386f3960a56c65516996cc/block.go#L24) message is like this 
```
message Block {
  bytes id
  bytes parent_id
  uint64 height
  google.protobuf.Timestamp timestamp
  repeated CollectionGuarantee collection_guarantees
  repeated BlockSeal block_seals
  repeated bytes signatures
}
```
Clearly `signatures` is missing in [block](https://github.com/onflow/flow-go-sdk/blob/573abf0bffa1334267386f3960a56c65516996cc/block.go#L24)

**Describe the solution you'd like**
Add signature field

PR #251, opened 2022-02-18
Implement signature field on Block struct

HarleyAppleChoi:
Closes: #250 

## Description

This implements the missing `signatures` field at [block](https://github.com/onflow/flow-go-sdk/blob/master/CONTRIBUTING.md#styleguides) struct at the FLOW api.

______

For contributor use:


test/entities.go
```diff
@@ -146,6 +149,7 @@ func (g *Blocks) New() *flow.Block {
 	return &flow.Block{
 		BlockHeader:  header,
 		BlockPayload: payload,
+		Signatures:   signiture,
```
sideninja, line 152:
```suggestion
		Signatures:   signature,
```

test/entities.go
```diff
@@ -146,6 +149,7 @@ func (g *Blocks) New() *flow.Block {
 	return &flow.Block{
```
sideninja, line 149: It would be good to actually add a test that checks the signature presence.

sideninja (approved):
Good find, thanks :)

test/entities.go
HarleyAppleChoi: I added the signature struct on test/entity.go and assertion on client/client.test.go Is that what you mean?

sideninja: Yes perfect, although I don't see the client test, did you push it?

HarleyAppleChoi: Sorry. My bad. I think client_test.go does not need to be updated.

sideninja: Good, then I would consider this done and approved. Thank you

Issue #246, opened 2022-02-09
`CalculateEventsHash` logic outdated


smnzhu:
It was brought to my attention that the logic in the SDK for computing Events Hashes is out of date with the protocol logic:

SDK:
https://github.com/onflow/flow-go-sdk/blob/573abf0bffa1334267386f3960a56c65516996cc/event.go#L94-L109

Flow:
https://github.com/onflow/flow-go/blob/60b67349445b054449ccef6cbd78644ec95b20c3/model/flow/event.go#L102-L124

I believe the protocol logic changed fairly recently, and it is now generating a Merkle tree rather than directly hashing the concatenation of the events. cc @ramtinms for specific details on the changes

ramtinms:
@sideninja I'm trying to evaluate a way to untangle this, currently talking to the team to see if we could move it to a separate package or part of the crypto lib

sideninja:
Does the interface has to keep the signature or it can change? I'm asking because the Merkle tree root hash is returned as identifiers and it might be a good idea to keep them as identifiers since they are more defined than hash.

sideninja:
DOD: 
- implemented
- released in new version of SDK

Current state: waiting for PR approvals cc @kerrywei

Darrenic17:
Connect with our  team representatives to help resolve your issue and troubleshooting .
Connect here: direct.lc.chat/14158209/

Hours of Operation
24/7 (Ready to fix issues)

PR #253, opened 2022-02-21
Calculating events hash using Patricia Merkle tree

sideninja:
Closes: #246 

## Description
Refactors how event hash is calculated. It now uses the flow-go Merkle Patricia tree for calculating the hash. 

A test example results in a change of hash:
```
      -0x0a1e6871892be16088661236ae90c3b3eb7d99e2a367faba55927425d000f43c (previous method)
      +0x7b3d0915bac54b34aba06160ec518ae3be4212a4dc18ad33163601e57f9b1e15 (current method)
```

______

For contributor use:


sideninja:
I would need a validation the implementation actually produces a valid output cc @ramtinms @smnzhu 
Since I can't really compare it to the previous result as it changed, it should right? Should I just write the same test in flow-go to test the output?

event.go
```diff
@@ -95,17 +97,52 @@ func (e *Event) Fingerprint() []byte {
...
+	if err != nil {
+		return nil, err
+	}
+
+	hash, err := hex.DecodeString(id.String())
+	if err != nil {
+		return nil, err
+	}
+
+	return hash, nil
```
ramtinms, line 115: not sure what this part is doing, if you need a crypto.Hash, this should work `crypto.Hash(id[:])`

sideninja: Mega facepalm! 🤦 🤦‍♂️ 🤦‍♀️ 
Thank you, fixed

sideninja:
Can we get this merged so it doesn't get outdated 🙏  cc @ramtinms @smnzhu

ramtinms (approved):

sideninja:
@psiemens can you help review this?

go.mod
```diff
@@ -3,16 +3,16 @@ module github.com/onflow/flow-go-sdk
...
-	cloud.google.com/go v0.97.0 // indirect
 	cloud.google.com/go/kms v1.0.0
-	github.com/ethereum/go-ethereum v1.9.9
-	github.com/onflow/cadence v0.20.1
-	github.com/onflow/flow-go/crypto v0.24.3
-	github.com/onflow/flow/protobuf/go/flow v0.2.2
+	github.com/ethereum/go-ethereum v1.9.13
+	github.com/onflow/cadence v0.21.2
+	github.com/onflow/flow-go v0.24.6
+	github.com/onflow/flow-go/crypto v0.21.4-0.20211125190211-7b31c986316e
```
psiemens, line 10: It looks like this downgrades the `flow-go/crypto` version. Is that intentional?

go.mod
```diff
@@ -3,16 +3,16 @@ module github.com/onflow/flow-go-sdk
...
 require (
-	cloud.google.com/go v0.97.0 // indirect
 	cloud.google.com/go/kms v1.0.0
-	github.com/ethereum/go-ethereum v1.9.9
-	github.com/onflow/cadence v0.20.1
-	github.com/onflow/flow-go/crypto v0.24.3
-	github.com/onflow/flow/protobuf/go/flow v0.2.2
+	github.com/ethereum/go-ethereum v1.9.13
+	github.com/onflow/cadence v0.21.2
+	github.com/onflow/flow-go v0.24.6
```
psiemens, line 9: I think we should avoid including `onflow/flow-go` as a dependency in `flow-go-sdk` because it will add a lot of bloat if we include the entire node software.

In this case, I would even say its OK to copy the implementation of `flow.EventsMerkleRootHash` to avoid having to import it.

psiemens: Ah, looks like it uses the Merkle tree package. Maybe that could be made into its own module?

sideninja: Yes, I remember having that exact conversation with Ramtin and we concluded that if it will work it's ok but he said is else willing to extract it somewhere. cc @ramtinms how should we proceed? I btw agree with this, but at the same time I also think we should also remove whole ethereum node module just for sake of RLP, so since that is now included it begs to ask how important this is, I remember it was required by some integration (not sure I can say which) and how long it will take to extract. Maybe Ramtin should say.

go.mod
tarakby: We should not downgrade the crypto version, if there is an issue with the newest crypto version, we can work together to fix it.

PR #258, opened 2022-03-24
Create new functions for GetTransactionResultByIndex and full block Responses

koko1123:
Closes: #2194

## Description
Create corresponding functions in go-sdk for the new GRPC endpoint

______

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -168,6 +168,25 @@ func (c *Client) GetLatestBlock(
 	return getBlockResult(res)
 }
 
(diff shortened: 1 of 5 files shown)


sideninja:
@koko1123 what is the extra data that is retrieved when fetching a full block? I'm asking that because I wonder whether that should simply become the default for the get block method unless the extra payload is pretty big. Otherwise, I feel we should also consider just adding a boolean to an existing API.

koko1123:
leaving as inactive, no longer needed

Issue #260, opened 2022-03-25
cloudkms signer is not threadsafe


judezhu:
## Instructions 

Please fill out the template below to the best of your ability and include a label indicating which tool/service you were working with when you encountered the problem. 

### Problem

The `hasher` of cloudkms signer returns error when running at a multi-thread env, because the underlying `sha256` lib was not thread safe. 

The error message is:

```
panic: d.nx != 0
```

The error was returned from [here](https://github.com/golang/go/blob/master/src/crypto/sha256/sha256.go#L234). 

### Steps to Reproduce 

We can reproduce this error by signing flow transactions in a multi-thread env:

```
	wg := sync.WaitGroup{}
	for i := 0; i < 2000; i++ {
		wg.Add(1)
		go func(wg *sync.WaitGroup, counter int) {
			defer wg.Done()
			tx := flow.NewTransaction().
				SetScript([]byte("")).
				SetGasLimit(9999).
				SetProposalKey(account.Address, key.Index, key.SequenceNumber).
				SetPayer(account.Address).
				AddAuthorizer(account.Address)

			if err := tx.SignEnvelope(account.Address, key.Index, signer); err != nil {
				log.Fatal(err)
			}
			log.Printf("[%d] done\n", counter)
		}(&wg, i)
	}

	wg.Wait()
```

The whole error stack is as below:
![image](https://user-images.githubusercontent.com/3269459/160212263-11a8e99b-2b78-45fe-87e9-449b2658306d.png)

PR #261, opened 2022-03-25
Fix cloudkms signer is not threadsafe

judezhu:
Closes: #260 

## Description

Create a new `hasher` in cloudkms signer, when signing flow transactions.

______

For contributor use:

### DIFF
--- a/crypto/cloudkms/signer.go
+++ b/crypto/cloudkms/signer.go
@@ -36,7 +36,6 @@ type Signer struct {
 	client   *kms.KeyManagementClient
 	key      Key
 	hashAlgo crypto.HashAlgorithm
-	hasher   crypto.Hasher
 }
 
 // SignerForKey returns a new Google Cloud KMS signer for an asymmetric key version.
@@ -49,17 +48,11 @@ func (c *Client) SignerForKey(
 		return nil, err
 	}
 
-	hasher, err := crypto.NewHasher(hashAlgo)
-	if err != nil {
-		return nil, fmt.Errorf("cloudkms: failed to instantiate hasher: %w", err)
-	}
-
 	return &Signer{
 		ctx:      ctx,
 		client:   c.client,
 		key:      key,
 		hashAlgo: hashAlgo,
-		hasher:   hasher,
 	}, nil
 }
 
@@ -67,7 +60,11 @@ func (c *Client) SignerForKey(
 //
 // Reference: https://cloud.google.com/kms/docs/create-validate-signatures
 func (s *Signer) Sign(message []byte) ([]byte, error) {
-	digest := s.hasher.ComputeHash(message)
+	hasher, err := crypto.NewHasher(s.hashAlgo)
+	if err != nil {
+		return nil, fmt.Errorf("cloudkms: failed to instantiate hasher: %w", err)
+	}
+	digest := hasher.ComputeHash(message)
 
 	digestMessage, err := makeDigest(s.hashAlgo, digest)
 	if err != nil {


turbolent (approved):
Thank you!

Alternatively we could put a mutex around it, or maybe just document that the API is not thread-safe and requires users to handle it. I'm OK with this solution

judezhu:
> Thank you!
> 
> Alternatively we could put a mutex around it, or maybe just document that the API is not thread-safe and requires users to handle it. I'm OK with this solution

Thanks for the quick feedback! We noticed this issue when we were trying to sign/send multiple flow transactions in parallel. We ended up with a workaround to get it working. But it will be much easier if we can support this in the upstream lib.

sideninja (approved):
Thank you, I think this makes sense, since moving the hash calculation is the only part making it unsafe, I don't believe there is much overhead doing so.

tarakby:
- it seems to me that the updated code is double-hashing the message to sign (once with the hasher + once with KMS). Did I miss something?
- Since there are still calls to hash on KMS (the same call that seems to cause the errors in the first place), I'm wondering if the issue is really solved from your tests?
- I didn't audit this code before and this was not added by this PR, but ideally the call to sign on KMS should include the hashing (this means the [request](https://github.com/judezhu/flow-go-sdk/blob/c5742b71d6a314193cba868f63f7f9b65844ab05/crypto/cloudkms/signer.go#L74) to sign should include the message and not the digest). Hopefully that would solve the thread issue too.

psiemens:
👍 to @tarakby's comment.

Cloud KMS actually allows you to either pass either un-hashed `data` or a pre-hashed `digest` (but not both), so I don't think this will produce the wrong signature. (ref: https://cloud.google.com/kms/docs/create-validate-signatures#data_versus_digest).

But I think we can just pass `data` and then just remove the hasher instance altogether.

tarakby:
> Cloud KMS actually allows you to either pass either un-hashed data or a pre-hashed digest (but not both), so I don't think this will produce the wrong signature

Thanks @psiemens, yes the kms interface allows to include either data or digest in the signing request. The original code looks correct to me, but I might be still missing something with the code updated after this PR: 
https://github.com/onflow/flow-go-sdk/blob/8fc9b5e084cf056ec09adec8004bfe33449dcc3b/crypto/cloudkms/signer.go#L63-L77

Line 67 computes the first hash with the hasher, line 69 computes the hash of the hash with KMS, line 76 includes the hash of the hash to be signed as a digest. Unless the verification is also double hashing, I guess this signature won't be valid.

tarakby:
In my comment I mainly mentioned 2 points:
 1. [correctness] double hashing should be avoided, otherwise it won't be validated by a correct verification.
 2. [best practice] if an HSM is used, it is recommended to delegate all cryptographic computations to it including hashing. Although hashing only deals with public data, it can lead to attacks in the context of signatures. An HSM is used when a the external environment isn't 100% trusted, so better to include hashing in the trusted environment too (in some designs that's not possible and that's why the kms API offers signing the digest).
 
 I'll open a PR to update 1 (pending your confirmation that I didn't misread the code) and 2.

psiemens:
@tarakby The call to `makeDigest` doesn't actually compute a hash, it just prepares the hash data structure required by KMS. So I don't think there is any double hashing.

But I agree with your point about delegating all cryptographic computations to KMS.

https://github.com/onflow/flow-go-sdk/blob/8fc9b5e084cf056ec09adec8004bfe33449dcc3b/crypto/cloudkms/signer.go#L92-L101

psiemens:
@tarakby A PR to fix point 2 would be great!

judezhu:
> double hashing should be avoided, otherwise it won't be validated by a correct verification.

I don't get the double hashing issue here. The change we made in this PR is not adding one more hash. Basically, we just replace the shared `hasher`, with a local `hasher` initialized in the `Sign` method.

tarakby:
> The call to makeDigest doesn't actually compute a hash, it just prepares the hash data structure required by KMS

My bad, I thought that function was computing a hash. Sorry @psiemens and @judezhu and thanks both for your clarifications!

PR #265, opened 2022-04-04
Update Cadence changes to memory metering

dsainati1:
Upcoming Cadence changes are adding another argument to `decode`, which normally is the memory gauge used to meter memory operations. Since these operations are client-side, however, we can just have this argument be `nil` here. 
______

For contributor use:

- [X] Targeted PR against `master` branch
- [X] Code follows the [standards mentioned here](https://github.com/onflow/flow-go-sdk/blob/master/CONTRIBUTING.md#styleguides).
- [X] Updated relevant documentation 
- [X] Re-reviewed `Files changed` in the Github PR explorer
- [X] Added appropriate labels


sideninja (commented):
Nice, thank you for this. The only flag is that you should also update the version of the cadence lib.

transaction.go
```diff
@@ -162,7 +162,7 @@ func (t *Transaction) Argument(i int) (cadence.Value, error) {
 
 	encodedArg := t.Arguments[i]
 
-	arg, err := jsoncdc.Decode(encodedArg)
+	arg, err := jsoncdc.Decode(nil, encodedArg)
```
turbolent, line 165: The CI error is weird: The function is called with 2 arguments, but it's (incorrectly?) reporting it's too many. However, in the test above, the function accepts 2 arguments. 🤔

dsainati1: This is because these changes require the secure cadence version, and CI is running on the latest actual release. Once we have an m4 release of Cadence we can update the go.mod here and CI should work.

turbolent: Ah, I see 👍

janezpodhostnik (approved):

Issue #193, opened 2021-05-05
Update usage of Google KMS API once secp256k1 keys are supported


psiemens:
GCP recently added support for the secp256k1 elliptic curve in Google KMS.

The option is available through the console UI, but there is not yet an enum value in their API.

This PR implements a temporary workaround to allow for signing via this SDK, but we should replace it once the KMS API has first party support: https://github.com/onflow/flow-go-sdk/pull/192

gatorbate32038:
// File: ./src/App.js import React from "react" import {AuthCluster} from "./auth-cluster" import {InitCluster} from "./init-cluster" import {ProfileCluster} from "./profile-cluster" import {useCurrentUser} from "./hooks/current-user" export default function App() { const cu = useCurrentUser() return ( <div> <AuthCluster /> <InitCluster address={cu.addr} /> <ProfileCluster address={cu.addr} /> <ProfileCluster address="0xba1132bc08f82fe2" /> <ProfileCluster address="0xf117a8efa34ffd58" /> </div> ) }

gatorbate32038:
// File: ./src/App.js import React from "react" import {AuthCluster} from "./auth-cluster" import {InitCluster} from "./init-cluster" import {ProfileCluster} from "./profile-cluster" import {useCurrentUser} from "./hooks/current-user" export default function App() { const cu = useCurrentUser() return ( <div> <AuthCluster /> <InitCluster address={cu.addr} /> <ProfileCluster address={cu.addr} /> <ProfileCluster address="0xba1132bc08f82fe2" /> <ProfileCluster address="0xf117a8efa34ffd58" /> </div> ) }

sideninja:
I believe this algos are not supported: https://pkg.go.dev/google.golang.org/genproto@v0.0.0-20220112215332-a9c7c0acf9f2/googleapis/cloud/kms/v1#CryptoKeyVersion_CryptoKeyVersionAlgorithm

sideninja:
It looks like they added the EC in enum https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/kms/v1#CryptoKeyVersion_CryptoKeyVersionAlgorithm

PR #266, opened 2022-04-04
[Crypto] general updates

tarakby:
## Description

- make changes to the kms signer `Sign` function so that hashing the message is done by the cloud KMS.
- add Keccak256 to the package supported algorithms.
- tidy up the ECDSA curves/hash functions compatibility. In particular, separate compatibility for Flow account keys and compatibility of using ECDSA.
- Closes: #193 (tidy up the support of secp256k1 by the kms signer).
- add a new `PublicKey` method to the `Signer` interface. 

______

For contributor use:


crypto/hash.go
```diff
@@ -64,3 +97,6 @@ func NewSHA3_256() Hasher {
 func NewSHA3_384() Hasher {
 	return hash.NewSHA3_384()
 }
+
+// NewKeccak_256 returns a new instance of Keccak256 hasher.
+var NewKeccak_256 = hash.NewKeccak_256
```
tarakby, line 102: I was wondering if I can update all the functions of type :
```
func SdkFunction() {
    flow_go_crypto.Funtion()
}
```
to 
```
var  SdkFunction = flow_go_crypto.Funtion
```
Is there a reason the first type was preferred so far?

crypto/crypto.go
```diff
@@ -53,31 +53,11 @@ func StringToSignatureAlgorithm(s string) SignatureAlgorithm {
...
-	case SHA3_256.String():
-		return SHA3_256
-
-	default:
-		return UnknownHashAlgorithm
-	}
-}
-
-// CompatibleAlgorithms returns true if the signature and hash algorithms are compatible.
+// CompatibleAlgorithms returns true if the signature and hash algorithms is a valid pair for a signing key
```
tarakby, line 56: After moving the hashing related stuff to `hash.go`, this function only contains ECDSA signing functions. I would like to rename the file to `sign.go`. I will proceed to rename it later to not complicate the review. Let me know if there is a concern with the renaming.

crypto/cloudkms/signer.go
```diff
@@ -22,113 +22,115 @@ import (
...
-
-	digestMessage, err := makeDigest(s.hashAlgo, digest)
-	if err != nil {
-		return nil, fmt.Errorf("cloudkms: failed to construct digest: %w", err)
-	}
 
 	request := &kmspb.AsymmetricSignRequest{
-		Name:   s.key.ResourceID(),
-		Digest: digestMessage,
+		Name:       s.key.ResourceID(),
```
tarakby, line 76: This change was not tested and requires testing with a KMS. How do you usually test these KMS changes ?  @psiemens @judezhu 
Ideally, the test would be added to CI (maybe have a kms key (ring) accessible by CI if that's simple) ?

account.go
```diff
@@ -95,15 +95,32 @@ func (a AccountKey) Encode() []byte {
...
+// accountCompatibleAlgorithms returns true if the signature and hash algorithms are a valid pair
+// for a key on a Flow account.
+func accountCompatibleAlgorithms(sigAlgo crypto.SignatureAlgorithm, hashAlgo crypto.HashAlgorithm) bool {
+	switch sigAlgo {
+	case crypto.ECDSA_P256:
+		fallthrough
+	case crypto.ECDSA_secp256k1:
+		switch hashAlgo {
+		case crypto.SHA2_256:
+			fallthrough
```
sideninja, line 107: nit: isn't this a bit confusing? although it's technically correct in this case it just replaces the `return true` in the next case and I wonder if that makes it better? I feel like it would be more clear if it would just return true.

account_test.go
```diff
@@ -77,18 +78,43 @@ func TestAccountKey(t *testing.T) {
...
+				crypto.SHA2_256: true,
+				crypto.SHA3_256: true,
+			},
+		}
 
-		key.SetHashAlgo(crypto.SHA3_256)
-		key.SetSigAlgo(crypto.UnknownSignatureAlgorithm)
-		assert.EqualError(t, key.Validate(), "signing algorithm (UNKNOWN) is incompatible with hashing algorithm (SHA3_256)")
+		key := AccountKey{}
+		for _, s := range signAlgos {
```
sideninja, line 108: nice 👍

crypto/cloudkms/signer.go
sideninja: I was testing that manually before. You need to have access to KMS (contact SRE) and then create a key (there should also be one already created by me a while ago). Like you said adding this to the CI and making an automatic test for it would be way better 💯

crypto/cloudkms/signer.go
```diff
@@ -22,113 +22,115 @@ import (
...
+// Only P-256 and secp256k1 are supported. The calling function should make sure
+// the function is only called with one of the 2 curves.
+func curveOrder(curve crypto.SignatureAlgorithm) int {
+	switch curve {
+	case crypto.ECDSA_P256:
+		return 32
+	case crypto.ECDSA_secp256k1:
+		return 32
+	default:
+		return 0 // or panic? this only happens if there is an implementation bug
```
sideninja, line 130: panic would probably be better

crypto/hash.go
sideninja: I don't know the initial design decisions (so hopefully Pete will speak to that) but this proposed change I guess has a downside of the possibility to overwrite that value saved there. Also, the call stack would be different but that I don't think has much of purpose anyway.

sideninja (approved):
Great improvements, thanks. LGTM 🚀

crypto/cloudkms/signer.go
judezhu: hey @tarakby, yeah! same here. In my local testing, we were using a `kms` key hosted on GCP. You probably can mock the `kms` sign function: `AsymmetricSign`, if you wanna get rid of the `kms` dependency in our local testing here.

LanfordCai:
I'm wondering maybe you will be interested in this: https://github.com/onflow/flow-go/issues/2251. It's also related to flow-go-sdk

tarakby:
This PR isn't completed (to address comments) and isn't fully tested yet (kms new signing request). I just didn't get time to get back to it recently but I should work on it next week. Is reverting the merge possible?

sideninja:
Yes, it is. Sorry for merging but if the PR is open and has approval I assume to merge it. Maybe we need to be more clear about who is also merging them in the future.

tarakby:
Agree, I should have said clearly that I'm adding more changes, or blocked merging. 
In `flow-go`, we are used to let the PR creator merge when they get approvals and think it's ready. I assumed the same for `flow-go-sdk`.

sideninja:
Yeah, no worries @tarakby it's also partially my fault for not clearing it with you. There's no protocol for go-sdk since there's only me actively maintaining but what you said makes sense.

crypto/hash.go
tarakby: Overwriting is a good point, though I think there is a low risk for a developer to overwrite a that value. 
For the call stack, since this is a short function, the complier might be already inlining it. 

I'll just keep the original code :)

sideninja: Honestly, I'm fine either way, I just wanted to provide my answer to the question.

account.go
tarakby: I totally agree 😄 the `fallthrough` was in the original code and I kept it. I'll update it.

PR #279, opened 2022-04-21
Add account key changes

janezpodhostnik:
## Description

Made AccountKeyToCadenceCryptoKey public so it can be used elsewhere as it is a very useful way to encode keys.

Changed newHashAlgoValue and  newSignAlgoValue because newHashAlgoValue was not able to encode some hash algos correclty.
______

For contributor use:

### DIFF
--- a/templates/accounts.go
+++ b/templates/accounts.go
@@ -114,7 +114,18 @@ func newPublicKeyValue(pubKey crypto.PublicKey) cadence.Struct {
 	)
 }
 
-func newKeyListValue(key *flow.AccountKey) cadence.Struct {
+// AccountKeyToCadenceCryptoKey converts a `flow.AccountKe` key to the Cadence struct `Crypto.KeyListEntry`,
+// so that it can more easily be used as a parameter in scripts and transactions.
+//
+// example:
+// ```go
+// 	key := AccountKeyToCadenceCryptoKey(accountKey)
+//
+//	return flow.NewTransaction().
+//		SetScript([]byte(templates.AddAccountKey)).
+//		AddRawArgument(jsoncdc.MustEncode(key))
+// ```
+func AccountKeyToCadenceCryptoKey(key *flow.AccountKey) cadence.Value {
 	weight, _ := cadence.NewUFix64(fmt.Sprintf("%d.0", key.Weight))
 
 	return cadence.NewStruct([]cadence.Value{
@@ -160,7 +171,7 @@ func CreateAccount(accountKeys []*flow.AccountKey, contracts []Contract, payer f
 	contractKeyPairs := make([]cadence.KeyValuePair, len(contracts))
 
 	for i, key := range accountKeys {
-		keyList[i] = newKeyListValue(key)
+		keyList[i] = AccountKeyToCadenceCryptoKey(key)
 	}
 
 	for i, contract := range contracts {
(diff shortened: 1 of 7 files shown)


templates/accounts.go
```diff
@@ -99,7 +114,18 @@ func newPublicKeyValue(pubKey crypto.PublicKey) cadence.Struct {
 	)
 }
 
-func newKeyListValue(key *flow.AccountKey) cadence.Struct {
+// AccountKeyToCadenceCryptoKey converts a `flow.AccountKe` key to the Cadence struct `Crypto.KeyListEntry`,
```
robert-e-davidson3, line 117:
```suggestion
// AccountKeyToCadenceCryptoKey converts a `flow.AccountKey` key to the Cadence struct `Crypto.KeyListEntry`,
```

templates/accounts.go
```diff
@@ -58,26 +57,42 @@ func exportType(t sema.Type) cadence.Type {
...
-	if sigAlgo == crypto.ECDSA_secp256k1 {
-		sigAlgoCadence = sema.SignatureAlgorithmECDSA_secp256k1
+	sigAlgoValue := sema.SignatureAlgorithmECDSA_P256
+	switch sigAlgo {
+	case crypto.ECDSA_P256:
+		sigAlgoValue = sema.SignatureAlgorithmECDSA_P256
+	case crypto.ECDSA_secp256k1:
+		sigAlgoValue = sema.SignatureAlgorithmECDSA_secp256k1
+	default:
+		panic(fmt.Sprintf("unsupported signature algorithm: %v", sigAlgo))
```
sideninja, line 67: should we return an error instead?

templates/accounts.go
```diff
@@ -58,26 +57,42 @@ func exportType(t sema.Type) cadence.Type {
...
 	}).WithType(
 		exportType(sema.SignatureAlgorithmType).(*cadence.EnumType),
 	)
 }
 
 func newHashAlgoValue(hashAlgo crypto.HashAlgorithm) cadence.Enum {
-	hashAlgoCadence := sema.HashAlgorithmSHA3_256
-	if hashAlgo == crypto.SHA2_256 {
-		hashAlgoCadence = sema.HashAlgorithmSHA2_256
+	hashAlgoValue := sema.HashAlgorithmSHA2_256
```
sideninja, line 78: 👍

templates/accounts.go
```diff
@@ -99,7 +114,18 @@ func newPublicKeyValue(pubKey crypto.PublicKey) cadence.Struct {
...
+//
+// example:
+// ```go
+// 	key := AccountKeyToCadenceCryptoKey(accountKey)
+//
+//	return flow.NewTransaction().
+//		SetScript([]byte(templates.AddAccountKey)).
+//		AddRawArgument(jsoncdc.MustEncode(key))
+// ```
+func AccountKeyToCadenceCryptoKey(key *flow.AccountKey) cadence.Value {
```
sideninja, line 128: 👌

sideninja (approved):
Thank you for better handling hashing algos. Good stuff.

templates/accounts.go
janezpodhostnik: we would have to propagate that error out of `AccountKeyToCadenceCryptoKey` then. Hm... yes, I guess It would be more proper. Let me see how big of a change this is.

sideninja: Yeah, is that used anyplace else? If it's too much of changes then maybe it can stay.

janezpodhostnik: I actually added it it in the emulator, e2e tests, flow-go and flow-core-contracts :). However for correctness sake I think it's better to change it.

robert-e-davidson3 (approved):

templates/accounts.go
janezpodhostnik: what do you think d263b74423232e9e10d88059dc86f4bfac8b02a0

sideninja (approved):

PR #283, opened 2022-05-06
Crypto: General clean up + KMS signing updates 

tarakby:
## Description

- update the KMS signer `Sign` function so that hashing is performed by KMS.
- add checksum check for signed data by KMS.
- add a manual test for KMS signing (skipped by CI)
- add Keccak256 to the package supported algorithms.
- tidy up the ECDSA curves/hash functions compatibility. In particular, separate compatibility for Flow account keys and compatibility of using ECDSA signature from the SDK.
- Closes: #193 (tidy up the support of secp256k1 by the kms signer).

______

For contributor use:


account.go
```diff
@@ -95,15 +95,26 @@ func (a AccountKey) Encode() []byte {
...
 
+// accountCompatibleAlgorithms returns true if the signature and hash algorithms are a valid pair
+// for a key on a Flow account.
+func accountCompatibleAlgorithms(sigAlgo crypto.SignatureAlgorithm, hashAlgo crypto.HashAlgorithm) bool {
+	if sigAlgo == crypto.ECDSA_P256 || sigAlgo == crypto.ECDSA_secp256k1 {
+		if hashAlgo == crypto.SHA2_256 || hashAlgo == crypto.SHA3_256 {
+			return true
+		}
+	}
+	return false
```
sideninja, line 106:
```suggestion
        return (sigAlgo == crypto.ECDSA_P256 || sigAlgo == crypto.ECDSA_secp256k1) && 
        (hashAlgo == crypto.SHA2_256 || hashAlgo == crypto.SHA3_256)
```

account_test.go
```diff
@@ -77,18 +78,43 @@ func TestAccountKey(t *testing.T) {
 		assert.EqualError(t, key.Validate(), "invalid key weight: -1")
 	})
 
-	t.Run("Invalid Key Algorithm", func(t *testing.T) {
-		privateKey := generateKey()
-		key := AccountKey{
-			PublicKey: privateKey.PublicKey(),
+	t.Run("Key Algorithm", func(t *testing.T) {
```
sideninja, line 81: ❤️

crypto/cloudkms/cloudkms_test.go
```diff
@@ -45,3 +51,84 @@ func TestKeyFromResourceID(t *testing.T) {
...
+	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", googleApplicationCreds)
+
+	return nil
+}
+
+// TestManualKMSSigning tests signing using a KMS key.
+// This tests requires access to KMS and cannot be run by CI.
+// Please use this test manually by commenting t.Skip(),
+// when making any change to the KMS signing code.
+func TestManualKMSSigning(t *testing.T) {
```
sideninja, line 92: Nice

crypto/crypto.go
```diff
@@ -119,8 +93,15 @@ type InMemorySigner struct {
 var _ Signer = (*InMemorySigner)(nil)
 
 // NewInMemorySigner initializes and returns a new in-memory signer with the provided private key
-// and hasher.
+// and hashing algorithm.
 func NewInMemorySigner(privateKey PrivateKey, hashAlgo HashAlgorithm) InMemorySigner {
+	// check compatibility to form a signing key
+	if !CompatibleAlgorithms(privateKey.Algorithm(), hashAlgo) {
+		// TODO: panic?
+		return InMemorySigner{}
```
sideninja, line 101: Should an error be returned here as well?

sideninja (approved):
Left some comments, but LGTM

account.go
tarakby: 👌🏼

crypto/crypto.go
tarakby: Good catch, I missed getting back to that. 
If this was an internal function I would add an error. For the public function, this will cause a breaking change. If you're okay with that I will add it. Any preference?

sideninja: I would say let us break it while we are not 1.0 after that it will be just harder. The problem if we don't do that we will just support "anti-patterns" and that's not good either.

crypto/cloudkms/cloudkms_test.go
```diff
@@ -45,3 +51,84 @@ func TestKeyFromResourceID(t *testing.T) {
...
+		ProjectID:  "dl-flow",
+		LocationID: "us-east1",
+		KeyRingID:  "testnet_keyring",
+		KeyID:      "sdk_test",
+		KeyVersion: "1",
+	}
+
+	// id is copied from the kms key: https://console.cloud.google.com/security/kms/key/manage/us-east1/testnet_keyring/sdk_test?project=dl-flow
+	// this test insures the key exists in the kms keyring
+	id := "projects/dl-flow/locations/us-east1/keyRings/testnet_keyring/cryptoKeys/sdk_test/cryptoKeyVersions/1"
```
psiemens, line 106: Can we make this an env var, like `GOOGLE_APPLICATION_CREDENTIALS`? I don't feel great about putting GCP details in code, even if they aren't technically secrets.

```go
id := os.Getenv(`KEY_RESOURCE_ID`)
key := cloudkms.KeyFromResourceID()
```

I'd then suggest we use a separate test to check that `key.ResourceID()` behaves as expected.

crypto/cloudkms/signer.go
```diff
@@ -64,77 +70,62 @@ func (c *Client) SignerForKey(
...
+	if _, err := asn1.Unmarshal(kmsSignature, &parsedSig); err != nil {
 		return nil, fmt.Errorf("asn1.Unmarshal: %w", err)
 	}
 
-	rBytes := parsedSig.R.Bytes()
-	rBytesPadded := rightPad(rBytes, ecCoupleComponentSize)
+	curveOrderLen := curveOrder(curve)
+	signature := make([]byte, 2*curveOrderLen)
 
+	// left pad R and S with zeroes
```
psiemens, line 108: So just to confirm: we want to left pad instead of right pad?

psiemens (commented):
Thanks for making these improvements!

crypto/crypto.go
tarakby: https://github.com/onflow/flow-go-sdk/pull/283/commits/7c53dc1d5babe18187ec5c1b63a34c31397df3dc

crypto/cloudkms/cloudkms_test.go
tarakby: Nice, I like your suggestion, added it in https://github.com/onflow/flow-go-sdk/pull/283/commits/cb01b7dc5629c7f3f6b3681c82631c7a4c01747d

crypto/cloudkms/signer.go
tarakby: if `(b_0, b_1, ..., b_n)` is the big endian representation of `R`, what we want to get is `[ 0, 0, .. 0, b_0, b_1, .., b_n]`. I assumed this is called left padding. Btw, the original code did it right. 

What I updated is to use the curve order length to pad, instead of the "curve size" or prime length (that is the 256 bits in P256 for instance). These are 2 different values, but they happened to be the same for P256. For a general curve, `R` and `S` should use the curve order.

PR #304, opened 2022-06-27
Update to Cadence 5a06fec82b4a

turbolent:
______

For contributor use:


sideninja:
@turbolent what is with the tests?

sideninja:
> > @turbolent what is with the tests?
> 
> Cadence is now using some features that only exist in Go 1.18, so to make this update work we are going to have to update the sdk to 1.18 as well.

Oh I see. Yeah that should be updated then. Should we create a PR or will you? I'm OOO now for a while so can't do that. If nothing else we should have an issue.

dsainati1:
> > > @turbolent what is with the tests?
> > 
> > 
> > Cadence is now using some features that only exist in Go 1.18, so to make this update work we are going to have to update the sdk to 1.18 as well.
> 
> Oh I see. Yeah that should be updated then. Should we create a PR or will you? I'm OOO now for a while so can't do that. If nothing else we should have an issue.

I was going to do it in this PR unless the version upgrade requires a lot of changes.

janezpodhostnik (approved):

PR #309, opened 2022-07-12
Fixing doc link for Transaction Signature

Tonix517:
Fixing a broken external doc link for "How Signatures Work in Flow"


janezpodhostnik (approved):
Good catch!

Issue #317, opened 2022-08-24
Signing payloads with a size greater than 64kb fails while using CloudKMS


pgebheim:
### Problem

When updating contracts, the whole contract source is converted to hex and then sent as part of a contract update transactions. The resulting payload that needs to be signed thus can grow larger than 64kb. This runs into issues when using CloudKMS because it has a 64kb plain text limit on its signing (and encryption/decryption) API endpoints.

The issue was introduced in the [May 5th commit](https://github.com/onflow/flow-go-sdk/commit/959016ef8960f9376fb0db87f2572cf7dcf163ef) where the data hashing was moved from being done in-memory to happening in the signer.

### Steps to Reproduce 

- Attempt to deploy a contract larger than 64kb.

### Acceptance Criteria

- Any sized contract should be deployable
  
### Context

We've worked around the issue by using the older version of flow-cli, so this is not immediately blocking.

tarakby:
Thanks for creating the issue. 
I suggest to fix the issue by keeping the HSM hashing for data of size smaller than the limit, and pre-hashing outside the HSM for data larger than the limit. This would keep the in-HSM hashing as the default method.

PR #318, opened 2022-08-29
[Crypto] fix KMS signing bug for long messages

tarakby:
Closes: https://github.com/onflow/flow-go-sdk/issues/317

## Description

Fixes a bug when messages signed by Google KMS are longer than 65536 bytes. 
If the message is longer than the KMS limit, pre-hash the message outside KMS. 

______

For contributor use:

### DIFF
--- a/crypto/cloudkms/signer.go
+++ b/crypto/cloudkms/signer.go
@@ -43,6 +43,8 @@ type Signer struct {
 	curve crypto.SignatureAlgorithm
 	// public key for easier access
 	publicKey crypto.PublicKey
+	// Hash algorithm associated to the KMS signing key
+	hashAlgo crypto.HashAlgorithm
 }
 
 // SignerForKey returns a new Google Cloud KMS signer for an asymmetric signing key version.
@@ -52,7 +54,7 @@ func (c *Client) SignerForKey(
 	ctx context.Context,
 	key Key,
 ) (*Signer, error) {
-	pk, _, err := c.GetPublicKey(ctx, key)
+	pk, hashAlgo, err := c.GetPublicKey(ctx, key)
 	if err != nil {
 		return nil, err
 	}
@@ -63,6 +65,7 @@ func (c *Client) SignerForKey(
 		key:       key,
 		curve:     pk.Algorithm(),
 		publicKey: pk,
+		hashAlgo:  hashAlgo,
 	}, nil
 }
 
@@ -71,10 +74,32 @@ func (c *Client) SignerForKey(
 // Reference: https://cloud.google.com/kms/docs/create-validate-signatures
 func (s *Signer) Sign(message []byte) ([]byte, error) {
 
-	request := &kmspb.AsymmetricSignRequest{
-		Name:       s.key.ResourceID(),
-		Data:       message,
-		DataCrc32C: checksum(message),
+	// Google KMS supports signing messages without pre-hashing
+	// up to 65536 bytes. Beyond that limit, messages must be
+	// prehashed outside KMS.
+	kmsPreHashLimit := 65536
+
+	var request *kmspb.AsymmetricSignRequest
+	if len(message) <= kmsPreHashLimit {
+		// hash within KMS
+		request = &kmspb.AsymmetricSignRequest{
+			Name:       s.key.ResourceID(),
+			Data:       message,
+			DataCrc32C: checksum(message),
+		}
+	} else {
+		// this is guaranteed to only return supported hash algos by KMS
+		hasher, err := crypto.NewHasher(s.hashAlgo)
+		if err != nil {
+			return nil, fmt.Errorf("cloudkms: failed to sign: %w", err)
+		}
+		// pre-hash outside KMS
+		hash := hasher.ComputeHash(message)
+		request = &kmspb.AsymmetricSignRequest{
+			Name:         s.key.ResourceID(),
+			Digest:       getDigest(s.hashAlgo, hash),
+			DigestCrc32C: checksum(hash),
+		}
 	}
 	result, err := s.client.AsymmetricSign(s.ctx, request)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


sideninja (approved):
Thank you! LGTM

bluesign:
> If the message is longer than the KMS limit, pre-hash the message outside KMS.

Sorry the naive question but why not always pre-hash ?

crypto/cloudkms/signer.go
```diff
@@ -71,10 +74,32 @@ func (c *Client) SignerForKey(
...
+	var request *kmspb.AsymmetricSignRequest
+	if len(message) <= kmsPreHashLimit {
+		// hash within KMS
+		request = &kmspb.AsymmetricSignRequest{
+			Name:       s.key.ResourceID(),
+			Data:       message,
+			DataCrc32C: checksum(message),
+		}
+	} else {
+		// this is guaranteed to only return supported hash algos by KMS
```
bluesign, line 91: Can you explain this comment? Why this guarantee is needed?

tarakby:
@bluesign 
Good question. Although hashing does not involve a private key, it is still a sensitive operation and an important step in signing (the security of the signature algorithm depends on the hash function security). It is therefore recommended to perform hashing within the boundaries of the trusted platform.
In this case the HSM is the trusted platform, and it is better to fit as much signing steps into it as possible. This helps reducing the attack windows. 
In some cases/designs, we may want to pre-hash outside the trusted hardware for different reasons, that's why some HSMs allow signing hashes.

bluesign:
@tarakby 

> Although hashing does not involve a private key, it is still a sensitive operation and an important step in signing. (the security of the signature algorithm depends on the hash function security).

If I understand correctly, hash is the "weakest link" here it means.

>  It is therefore recommended to perform hashing within the boundaries of the trusted platform. In this case the HSM is the trusted platform, and it is better to fit as much signing steps into it as possible. This helps reducing the attack windows.

Here I think we differ. In Flow use case, other use cases I am not sure, only attack is to confuse Transaction with Hash. 

So I specially craft a Transaction, which looks innocent, but transaction canonical form will be equal to the hash of my evil transaction. 

But in this case minimum Flow Transaction Payload length, transaction domain tag etc all prevent those.

tarakby:
Regarding SHA3 support: In theory, it is possible to pre-hash using SHA3 and only process the following steps on HSM as you mentioned. For now we restricted the support to SHA2 for consistency because the ECDSA keys generated on KMS [specify the hashing algo](https://cloud.google.com/kms/docs/reference/rpc/google.cloud.kms.v1#cryptokeyversionalgorithm) and for now only include SHA2.

bluesign:
I was thinking more on what is feeling not ok for me, I think I figured it out. ( After a lot of coffee ) 

Consider a system with flow-go-sdk: ( let's say a web REST service for signing getting arbitrary transaction data )

User - Service - KMS - Flow Network:  let's say service is super dummy and just passing data to KMS. ( or let's say simply exposing KMS methods to web ) 

We have 1 method: `hashAndSign(m []byte)`, then hashAndSign(m []byte) is only vulnerable in the case of hash collision or bad implementation on hashing algorithm. 

Collision case ( on SHA ) can be ignored as this is not our concern ( as it will be similar bad in KMS as ours ), that leaves us with bad implementation on hashing algorithm. (causing a prefix collision attack, preimage attack etc ) 

in case of pre-image attack:
this is very unlikely ( as we validate it is a transaction etc ) But it is game over: if hash(evil_transaction) = kms_hash(innocent_transaction) So KMS doesn't make difference in this case ( considering we use the same hash algorithm in software of flow-go ) 
 
In case of prefix collision attack:
If we have bad hashing algorithm, and google has the correct, worse case is that we will reject valid signatures. 

So sending KSM here is obviously has benefits. ( for prefix collision attack) 

But now, when we add the size limit (65536), aren't we back to square one ?

Considering hashing algorithm is buggy and vulnerable to preimage attack,  as we can choose p1 and p2 in preimage attack, and only search for m1 and m2, we can keep both payloads over 65536 easily. 


PS:  I think we can most likely say that Transaction structure is not vulnerable to the preimage attack but it is not relevant in this discussion.

tarakby:
I see what kind of issues you were considering. 
Here is a high level of the situation as I see it:

> We have 1 method: hashAndSign(m []byte)

I would consider the method as `sign(m []byte)` which includes hashing as an internal step (a step that does not involve the private key). Here we're discussing whether `sign` would be split over two platforms. 

Reusing the same flow you described:  User - Service - KMS - Network, we should first assume the network (the party that verifies signatures) is using a secure hash function `hash_network` (both algorithm and implementation). Otherwise, multiple attack paths would open up, whatever signing implementation we use. `hash_network` is resistant to 1st pre-image and collisions (therefore 2nd pre-image too). 

Now we're comparing:
1. hash the message on service - send hash to KMS - finish signing on KMS
2. send message to KMS - sign on KMS

In both options, KMS does not check the data it receives corresponds to a valid transaction. The service is the party that is supposed to guarantee that.  Let's call `h_user` the hash output signed by the user (either in (1) or (2))
There will be an issue in general if  `h_user = hash_network(malicious_transaction)` i.e, the user authorizing `malicious_transaction` implicitly by signing `h_user`. 

We can also make the assumption that KMS implements the same `hash_network`. Signing `malicious_transaction` can only happen if something goes wrong at the service level :
 - in (1), the service is tricked to replace message by `malicious_transaction`, or the service is tricked to send `hash_network(malicious_transaction)` to KMS (using a faulty hash function on service or any other method).  
 - in (2),  the service is tricked to replace message by `malicious_transaction`. 
 
 I think there is no need in this case to consider the collision properties of `hash_service`. It is enough to get KMS to sign `hash_network(malicious_transaction)`. 
 
 (2) has less windows than (1) to trick the service, as it reduces the attack window to changing `message`.  Since it is not possible to implement (2) for long messages, it is fine to fall back to (1) in that case.

tarakby:
FYI, it seems to me that AWS HSM interface does not allow pre-hashing when signing (had a quick look at the API so I might be wrong) https://docs.aws.amazon.com/cloudhsm/latest/userguide/key_mgmt_util-sign.html

pgebheim:
> FYI, it seems to me that AWS HSM interface does not allow pre-hashing when signing (had a quick look at the API so I might be wrong) https://docs.aws.amazon.com/cloudhsm/latest/userguide/key_mgmt_util-sign.html

Iirc the hsm also includes a tighter limit on payload size.

bluesign:
> FYI, it seems to me that AWS HSM interface does not allow pre-hashing when signing (had a quick look at the API so I might be wrong) https://docs.aws.amazon.com/cloudhsm/latest/userguide/key_mgmt_util-sign.html

https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-mechanisms.html#mech3-2

> [3.2] AWS CloudHSM approaches hashing differently based on the Client SDK. For Client SDK 5, we do all hashing for supported functions in software and we do not copy to the HSM, meaning that there is no data size limit. For Client SDK 3, where we do the hashing depends on data size and whether you’re using single-part or multipart operations.

@pgebheim correct they seem to do 16kb max

PR #319, opened 2022-09-06
Remove deprecated GH action

10thfloor:
This action is no longer required to refresh doc content when files are updated.


turbolent:
What is the new way of updating documentation content?

sideninja (approved):

Issue #378, opened 2023-04-13
Add support for new event streaming API


nvdtf:
Flow team is working on delivering event streaming API: https://github.com/onflow/flips/pull/73/files (proto https://github.com/onflow/flow/pull/1275/files). 

This feature needs to be integrated into Flow Go SDK so all Go developers can also use it.

Example implementation for execution data streaming: https://github.com/peterargue/execdata-client/

PR #417, opened 2023-06-28
Add support for event streaming API

nvdtf:
Closes: #378 

Note that this feature is still in development in access nodes. I'll update the examples before merging when it becomes available on all nodes.


access/client.go
```diff
@@ -102,6 +102,15 @@ type Client interface {
 	// GetExecutionResultForBlockID gets the execution results at the block ID.
 	GetExecutionResultForBlockID(ctx context.Context, blockID flow.Identifier) (*flow.ExecutionResult, error)
 
+	// GetExecutionDataByBlockID returns execution data for a specific block ID.
+	GetExecutionDataByBlockID(ctx context.Context, blockID flow.Identifier) (*flow.ExecutionData, error)
+
+	// SubscribeExecutionData subscribes to execution data updates starting at the given block ID or height.
+	SubscribeExecutionData(ctx context.Context, startBlockID flow.Identifier, startHeight uint64) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error)
```
nvdtf, line 109: Might be a good idea to break this into two functions: `ByHeight` and `ById`

access/grpc/convert.go
```diff
@@ -542,3 +542,107 @@ func messageToTransactionResult(m *access.TransactionResultResponse, options []j
...
+
+	for i, ev := range m.GetEvents() {
+		res, err := messageToEvent(ev, nil)
+		if err != nil {
+			return nil, err
+		}
+		events[i] = &res
+	}
+
+	return &flow.ChunkExecutionData{
```
peterargue, line 595: this was recently updated to include transaction results

https://github.com/onflow/flow-go/blob/master/engine/common/rpc/convert/execution_data.go#L139

access/grpc/grpc.go
```diff
@@ -645,3 +657,149 @@ func (c *BaseClient) GetExecutionResultForBlockID(ctx context.Context, blockID f
...
+		defer close(sub)
+		defer close(errChan)
+
+		for {
+			resp, err := stream.Recv()
+			if err == io.EOF {
+				return
+			}
+			if err != nil {
+				errChan <- fmt.Errorf("error receiving execution data: %w", err)
```
peterargue, line 715: since we have a context, it's possible the caller stopped monitoring the streams. It's probably a good idea to add a `select` for each of these channel operations

access/grpc/grpc.go
```diff
@@ -645,3 +657,149 @@ func (c *BaseClient) GetExecutionResultForBlockID(ctx context.Context, blockID f
...
+				continue
```
peterargue, line 716: should this return or are stream errors recoverable?

access/grpc/grpc.go
```diff
@@ -645,3 +657,149 @@ func (c *BaseClient) GetExecutionResultForBlockID(ctx context.Context, blockID f
...
+			}
+
+			execData, err := messageToBlockExecutionData(resp.GetBlockExecutionData())
+			if err != nil {
+				errChan <- fmt.Errorf("error converting execution data: %w", err)
+				return
+			}
+
+			//log.Printf("received execution data for block %d %x with %d chunks", resp.BlockHeight, execData.BlockID, len(execData.ChunkExecutionDatas))
+
```
peterargue, line 726:
```suggestion
```

access/grpc/grpc.go
```diff
@@ -645,3 +657,149 @@ func (c *BaseClient) GetExecutionResultForBlockID(ctx context.Context, blockID f
...
+				errChan <- fmt.Errorf("error receiving execution data: %w", err)
+				continue
+			}
+
+			events := make([]flow.Event, len(resp.GetEvents()))
+			for _, ev := range resp.GetEvents() {
+				res, err := messageToEvent(ev, nil)
+				if err != nil {
+					errChan <- err
+					continue
```
peterargue, line 790: I don't think this error case should continue. If the message fails to decode, skipping here may result in missed events

access/grpc/grpc.go
```diff
@@ -645,3 +657,149 @@ func (c *BaseClient) GetExecutionResultForBlockID(ctx context.Context, blockID f
...
+		defer close(sub)
+		defer close(errChan)
+
+		for {
+			resp, err := stream.Recv()
+			if err == io.EOF {
+				return
+			}
+			if err != nil {
+				errChan <- fmt.Errorf("error receiving execution data: %w", err)
```
peterargue, line 781: same comment here about the `selects`

access/client.go
peterargue: @sideninja @nvdtf what's your opinion on this? Currently the rpc accepts either height/blockID which is a bit of an outlier for the Access API. All of the existing methods match the rpc.

1. Keep it consistent with the rpc (keep it as-is in this PR)
2. Keep it consistent with the other methods (add `*ByBlockID`, `*ByHeight` versions
3. Align the protobuf with 2. If we go this route, I vote we do 1, then make those changes separately.

nvdtf: I think 2 is good enough. To me, accepting two fields and relying on either one is not good developer experience. Go SDK as the SDK can break this into two functions and call the RPC properly. 
I like 3 as well, but don't want to push extra work for minor improvements.

peterargue: updated in [4275139](https://github.com/onflow/flow-go-sdk/pull/417/commits/4275139b17fea0894178a181979c1a83cc4022f4)

peterargue:
@sideninja, @nvdtf mind taking a look? once this is ready, I can take a stab at adding it into the flow-cli

sideninja (approved):

jribbink (approved):

PR #534, opened 2023-12-08
Crypto module update and BLS support

tarakby:
This PR imports the new `v0.25.0` version of the `onflow/crypto`. 

This new version supports the BLS signature algorithm without the need to specify any Go build tag or run a pre-build step. Therefore, it makes sense to support BLS in `flow-go-sdk/crypto` without development friction.
This allows developers to play with BLS signing since it is already supported by Cadence.

In this PR:
- update the crypto module version
- add BLS signature algorithm
- add standard KMAC 128 which can be used with BLS if set properly
- add a special hasher (KMAC based) that is pre-set to work with BLS properly and abstracts the complexities of configuring a hasher for BLS.
- update compatibility checks of signing/hashing algorithms and tests
- update Makefile by adding a crypto-related flag to `go build`, that is needed for old CPUs not supporting ADX instructions.
- update README with a cgo note

side change:
 - update cloudkms package after deprecation


Makefile
```diff
@@ -1,9 +1,14 @@
...
 
+# include script to possibly set a crypto flag for older machines
+include crypto_adx_flag.mk
+
+CGO_FLAG := CGO_CFLAGS=$(CRYPTO_FLAG)
+
 .PHONY: test
 test:
-	GO111MODULE=on go test -coverprofile=coverage.txt ./...
+	GO111MODULE=on $(CGO_FLAG) go test -coverprofile=coverage.txt ./...
```
tarakby, line 11: This flag basically disables fast code using ADX instructions. 
If the machine is old, that is necessary. If the machine is new, either way works, we just get slower times with a set flag. 

If you think the `.mk` file is disruptive, we can also remove it and force the build to always run without the fast code. This would work for all machines but always give slower times (for BLS operations only, so not a big impact)

sideninja (approved):

tarakby:
@gregor will this PR create any cgo build issue in another repo importing flow-go-sdk ?

PR #636, opened 2024-04-26
add backwards compatible decoder into grpc and http client

ianthpun:
Closes: #637

## Description

Create new `ClientOption` that wraps the grpc.DialOption and cadence json option so that the user can add those to their NewClient call
______

For contributor use:


access/grpc/client.go
```diff
@@ -42,20 +43,50 @@ const CanarynetHost = "access.canary.nodes.onflow.org:9000"
...
+
+func DefaultClientOptions() *options {
+	return &options{
+		dialOptions: []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())},
+		jsonOptions: []cdcjson.Option{cdcjson.WithAllowUnstructuredStaticTypes(true)},
+	}
+}
+
+// WithGRPCDialOptions wraps a grpc.DialOption into a ClientOption.
+func WithGRPCDialOptions(dialOpts ...grpc.DialOption) ClientOption {
```
ianthpun, line 62: these are modelled after SubscribeOptions, seen here
https://github.com/onflow/flow-go-sdk/blob/e56a70a692d1a6507ccc023032f67159ee923abc/access/grpc/grpc.go#L50-L73

access/grpc/client.go
```diff
@@ -42,20 +43,50 @@ const CanarynetHost = "access.canary.nodes.onflow.org:9000"
...
+
+type options struct {
+	dialOptions []grpc.DialOption
+	jsonOptions []cdcjson.Option
+}
+
+func DefaultClientOptions() *options {
+	return &options{
+		dialOptions: []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())},
+		jsonOptions: []cdcjson.Option{cdcjson.WithAllowUnstructuredStaticTypes(true)},
```
turbolent, line 57: Maybe add some newlines to make this more readable. Same below

```suggestion
		dialOptions: []grpc.DialOption{
			grpc.WithTransportCredentials(insecure.NewCredentials()),
		},
		jsonOptions: []cdcjson.Option{
			cdcjson.WithAllowUnstructuredStaticTypes(true),
		},
```

access/grpc/client.go
```diff
@@ -28,6 +28,7 @@ package grpc
 import (
 	"context"
 
+	cdcjson "github.com/onflow/cadence/encoding/json"
```
turbolent, line 31: Small nit: The format is usually referred to as "JSON-Cadence" or "JSON-CDC" and import aliases are usually `jsoncdc`

access/grpc/client.go
turbolent: This is great!

turbolent (commented):
Good idea, looks great!

Could you also please add some tests?

ianthpun:
@turbolent i've added tests to cover both, but the tests unforauntely cannot be done with a `assert.Contains` or `assert.Equal`, largely because its comparing arrays that hosts functions, and I have not been successful in doing a real assertion on them together, with really sad errors like

```
	Error:      	[]json.Option{(json.Option)(0x1017956f0), (json.Option)(0x1017ba6d0)} does not contain (json.Option)(0x1017ba6d0)
	```

I believe the tests are siginificant enough to know that they are working as intended

**Comment by turbolent** [comment]
@ianthpun Ah, yeah, I should have been more specific: Testing the options are set isn't all that useful, but it would be great to have an integration test that ensures that API responses with values encoded in the pre-1.0 format can be decoded successfully if the backward-compatibility option is provided. 

That might require quite a setup (mocks?), and might be too much work though.

**Comment by ianthpun** [comment]
@turbolent i thought about that as well, but I opted not to go with it as the amount of work to prove the integration would be much higher than simply to release it. Reason being, we already have tests no the `cadence` side that test against the decoder functionality, so having an integration test on the flow-go-sdk, especially that that would require us to _build_ the integration suite since its not availale, with it seems overkill. 

What are your thoughts on releasing this as-is? @turbolent

**Review by turbolent — APPROVED** [review]

**Comment by turbolent** [comment]
Sounds good!

File: access/grpc/client.go [hunk 4]
```diff
@@ -42,26 +43,60 @@ const CanarynetHost = "access.canary.nodes.onflow.org:9000"
...
+	client.SetJSONOptions(cfg.jsonOptions)
+
+	return &Client{grpc: client, dialOptions: cfg.dialOptions}, nil
 }
 
 // Client implements all common gRPC methods providing a network agnostic API.
 type Client struct {
-	grpc *BaseClient
+	grpc        *BaseClient
+	dialOptions []grpc.DialOption
```
jribbink, line 99: Is this necessary?

access/grpc/client.go
```diff
@@ -42,26 +43,60 @@ const CanarynetHost = "access.canary.nodes.onflow.org:9000"
...
+	jsonOptions []jsoncdc.Option
+}
+
+func DefaultClientOptions() *options {
+	return &options{
+		dialOptions: []grpc.DialOption{
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+		},
+		jsonOptions: []jsoncdc.Option{
+			jsoncdc.WithAllowUnstructuredStaticTypes(true)},
```
jribbink, line 60:
```suggestion
			jsoncdc.WithAllowUnstructuredStaticTypes(true),
		},
```

jribbink (approved):

Issue #666, opened 2024-05-23
Backwards Compatibility - Failed to get Transaction Result


austinkline:
### Problem

Getting a failure while trying to parse transactions results using the crescendo preview build's backwards compatibility option

```
failed to get transactions by block id: client: failed to construct flow.TransactionResult entity from protobuf value: ccf convert: ccf: failed to decode: unexpected CBOR type CBOR boolean type as Authorization type"
```

### Steps to Reproduce 

[This gist reproduces the failure](https://gist.github.com/austinkline/ec49921aa0b74790fea85c469e6c987a)

### Acceptance Criteria

This transaction should decode correctly

peterargue:
Hi @austinkline. the SDK uses CCF encoded events, so your update to use json backwards compatibility mode is not registered. We're looking into how to address this for CCF.

As a temporary workaround, you could use grpc directly to get the tx result, and leave the encoding version flag as its default (jsoncdc)

austinkline:
Would there maybe be a way to switch to using JSON encoding instead of ccf temporarily? I'm not in control of that I don't think and am just using standard sdk methods

peterargue:
it's currently not configurable. it's a simple change to switch it to jsoncdc if you wanted to try creating a custom build as a temporary workaround. just need to remove this line
https://github.com/onflow/flow-go-sdk/blob/master/access/grpc/grpc.go#L383

peterargue:
@austinkline I just tagged `v1.0.0-preview.33` which includes a change to set the encoding.

you can add the following to your client setup
```golang
	opts = append(opts, flowGrpc.WithEventEncoding(flow.EventEncodingVersionJSONCDC))
```

give that a try and let me know how it goes.

austinkline:
Amazing! I'll give it a try tomorrow and report back

peterargue:
event encoding option is live since the v1.0.0

PR #673, opened 2024-05-30
Make event encoding configurable over grpc

peterargue:
Closes: #666 

## Description

Currently, event encoding is hard coded to use CCF. This was added to switch to the smaller size payloads. However, it does not support backwards compatibility mode needed by some clients during the Cadence 1.0 transition.

This PR updates the grpc client code to allow setting the event encoding for the client. During the transition, users can switch to use JSONCDC.

______

For contributor use:

### DIFF
--- a/access/grpc/client.go
+++ b/access/grpc/client.go
@@ -49,8 +49,9 @@ const PreviewnetHost = "access.previewnet.nodes.onflow.org:9000"
 type ClientOption func(*options)
 
 type options struct {
-	dialOptions []grpc.DialOption
-	jsonOptions []jsoncdc.Option
+	dialOptions   []grpc.DialOption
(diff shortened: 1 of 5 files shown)


nvdtf (approved):
Nice!

access/grpc/client.go
```diff
@@ -103,6 +113,11 @@ type Client struct {
 	grpc *BaseClient
 }
 
+// RPCClient returns the underlying gRPC client.
+func (c *Client) RPCClient() RPCClient {
```
peterargue, line 117: this isn't strictly needed here, but it provides some flexibility for users in the future to break out into a raw grpc request when the sdk doesn't expose some needed/new functionality

turbolent (commented):
Nice!

PR #738, opened 2024-09-03
Add fee method on transaction results

sideninja:
Add a `Fee()` method on the transaction result, which will return the Flow fee used in transaction execution.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/onflow/go-ethereum/rlp"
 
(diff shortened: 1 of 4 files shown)


transaction.go
```diff
@@ -627,6 +628,31 @@ type TransactionResult struct {
...
+// the transaction result.
+func (t *TransactionResult) Fee() (uint64, error) {
+	var feeEvent cadence.Event
+	for _, e := range t.Events {
+		if strings.Contains(e.Type, "FlowFees.FeesDeducted") {
+			if feeEvent.Type() != nil {
+				return 0, fmt.Errorf("could not extract transaction fee")
+			}
+			feeEvent = e.Value
+		}
```
janezpodhostnik, line 644:
```suggestion
			feeEvent = e.Value
			break
		}
```

transaction.go
```diff
@@ -627,6 +628,31 @@ type TransactionResult struct {
...
+	}
+
```
janezpodhostnik, line 646: fees could be disabled
```suggestion
if feeEvent == nil {
    return 0, nil
}
```

janezpodhostnik (changes requested):
Convenient!

nvdtf (approved):

Issue #748, opened 2024-09-11
Add GetCollectionByID and GetFullCollectionByID endpoints


Guitarheroua:
## Instructions 

Add `GetCollectionByID` and `GetFullCollectionByID` to the SDK that were implemented for Access API.

PR #754, opened 2024-09-12
Add GetFullCollectionById endpoint

illia-malachyn:
Closes: #748

### DIFF
--- a/access/client.go
+++ b/access/client.go
@@ -62,7 +62,7 @@ type Client interface {
 	GetBlockByHeight(ctx context.Context, height uint64) (*flow.Block, error)
 
 	// GetCollection gets a collection by ID.
-	GetCollection(ctx context.Context, colID flow.Identifier) (*flow.LightCollection, error)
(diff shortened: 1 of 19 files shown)


access/client.go
```diff
@@ -62,7 +62,7 @@ type Client interface {
 	GetBlockByHeight(ctx context.Context, height uint64) (*flow.Block, error)
 
 	// GetCollection gets a collection by ID.
-	GetCollection(ctx context.Context, colID flow.Identifier) (*flow.Collection, error)
+	GetCollection(ctx context.Context, colID flow.Identifier) (*flow.LightCollection, error)
```
illia-malachyn, line 65: I'm not sure if it is a good idea to change type's name as it might break code that depends on this. 
However, we want to do this to differentiate between light & full collection.

illia-malachyn: reverted renaming

nvdtf (approved):

Issue #746, opened 2024-09-11
Add SubscribeBlocks*, SubscribeBlockHeaders*, and SubscribeBlockDigests* endpoints


Guitarheroua:
## Instructions 

Add next endpoints to the SDK that were implemented for Access API:

* `SubscribeBlocksFromStartBlockID`
* `SubscribeBlocksFromStartHeight`
* `SubscribeBlocksFromLatest`
* `SubscribeBlockHeadersFromStartBlockID`
* `SubscribeBlockHeadersFromStartHeight`
* `SubscribeBlockHeadersFromLatest`
* `SubscribeBlockDigestsFromStartBlockID`
* `SubscribeBlockDigestsFromStartHeight`
* `SubscribeBlockDigestsFromLatest`

PR #758, opened 2024-09-18
Add subscribe blocks endpoints

illia-malachyn:
Part of #746

### DIFF
--- a/access/grpc/client.go
+++ b/access/grpc/client.go
@@ -165,6 +165,14 @@ func (c *Client) GetCollection(ctx context.Context, colID flow.Identifier) (*flo
 	return c.grpc.GetCollection(ctx, colID)
 }
 
+func (c *Client) GetCollectionByID(ctx context.Context, id flow.Identifier) (*flow.Collection, error) {
+	return c.grpc.GetLightCollectionByID(ctx, id)
+}
+
+func (c *Client) GetFullCollectionByID(ctx context.Context, id flow.Identifier) (*flow.FullCollection, error) {
+	return c.grpc.GetFullCollectionByID(ctx, id)
+}
+
 func (c *Client) SendTransaction(ctx context.Context, tx flow.Transaction) error {
 	return c.grpc.SendTransaction(ctx, tx)
 }
--- a/access/grpc/client.go
+++ b/access/grpc/client.go
@@ -173,8 +173,16 @@ func (c *Client) GetTransaction(ctx context.Context, txID flow.Identifier) (*flo
 	return c.grpc.GetTransaction(ctx, txID)
 }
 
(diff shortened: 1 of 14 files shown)


janezpodhostnik (commented):
There is a lot of repeated boilerplate happening here.

How about using a generic method like this
```go
func genericSubscribe[Receive, Response any](
	ctx context.Context,
	receive func() (Receive, error),
	convert func(Receive) (Response, error),
	topicNameForErrors string,
) (<-chan Response, <-chan error) {
	sub := make(chan Response)
	errChan := make(chan error)

	sendErr := func(err error) {
		select {
		case <-ctx.Done():
		case errChan <- err:
		}
	}

	go func() {
		defer close(sub)
		defer close(errChan)

		for {
			resp, err := receive()
			if err != nil {
				if err == io.EOF {
					return
				}

				sendErr(fmt.Errorf("error receiving %s: %w", topicNameForErrors, err))
				return
			}

			response, err := convert(resp)
			if err != nil {
				sendErr(fmt.Errorf("error converting %s: %w", topicNameForErrors, err))
				return
			}

			select {
			case <-ctx.Done():
				return
			case sub <- response:
			}
		}
	}()

	return sub, errChan
}
```

Using it would look like this in the case of `subscribeExecutionData`:

```go

func (c *BaseClient) subscribeExecutionData(
	ctx context.Context,
	req *executiondata.SubscribeExecutionDataRequest,
	opts ...grpc.CallOption,
) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error) {
	stream, err := c.executionDataClient.SubscribeExecutionData(ctx, req, opts...)
	if err != nil {
		return nil, nil, err
	}

	sub, errChan := genericSubscribe(
		ctx,
		stream.Recv,
		func(resp *executiondata.SubscribeExecutionDataResponse) (flow.ExecutionDataStreamResponse, error) {
			execData, err := convert.MessageToBlockExecutionData(resp.GetBlockExecutionData())
			if err != nil {
				return flow.ExecutionDataStreamResponse{}, fmt.Errorf("error converting execution data for block %d: %w", resp.GetBlockHeight(), err)
			}

			response := flow.ExecutionDataStreamResponse{
				Height:         resp.BlockHeight,
				ExecutionData:  execData,
				BlockTimestamp: resp.BlockTimestamp.AsTime(),
			}
			return response, nil
		},
		"execution data",
	)

	return sub, errChan, nil
}
```

Another question I have is how do we feel about changing the return type to `(<-chan flow.ExecutionDataStreamResponse, <-chan error)` from `(<-chan flow.ExecutionDataStreamResponse, <-chan error, error) `  by simply sending that first error on the error channel. This might be easier to handle.

franklywatson:
@illia-malachyn sounds good thanks

peterargue (approved):

peterargue:
@illia-malachyn can you merge master. after that this should be ready to merge

Issue #747, opened 2024-09-11
 Add SubscribeAccountStatuses* endpoints


Guitarheroua:
## Instructions 

Add next endpoints to the SDK that was implemented for Access API:

* `SubscribeAccountStatusesFromStartBlockID`
* `SubscribeAccountStatusesFromStartHeight`
* `SubscribeAccountStatusesFromLatestBlock`

PR #762, opened 2024-09-20
Add subscribe account statuses endpoint

illia-malachyn:
Closes: #747

### DIFF
--- a/access/grpc/grpc.go
+++ b/access/grpc/grpc.go
@@ -23,6 +23,7 @@ package grpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"sync/atomic"
@@ -1188,7 +1189,7 @@ func (c *BaseClient) SubscribeAccountStatusesFromStartHeight(
 	go func() {
 		defer close(accountStatutesChan)
 		defer close(errChan)
-		receiveAccountStatusesFromClient(ctx, subscribeClient, accountStatutesChan, errChan, c.msgIndex)
+		receiveAccountStatusesFromClient(ctx, subscribeClient, accountStatutesChan, errChan, &c.msgIndex)
 	}()
 
 	return accountStatutesChan, errChan, nil
@@ -1220,7 +1221,7 @@ func (c *BaseClient) SubscribeAccountStatusesFromStartBlockID(
 	go func() {
 		defer close(accountStatutesChan)
 		defer close(errChan)
-		receiveAccountStatusesFromClient(ctx, subscribeClient, accountStatutesChan, errChan, c.msgIndex)
+		receiveAccountStatusesFromClient(ctx, subscribeClient, accountStatutesChan, errChan, &c.msgIndex)
 	}()
 
 	return accountStatutesChan, errChan, nil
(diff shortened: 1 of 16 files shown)


janezpodhostnik (approved):

access/grpc/grpc.go
```diff
@@ -1129,3 +1129,138 @@ func (c *BaseClient) subscribeEvents(
...
+			sendErr(fmt.Errorf("error receiving account status: %w", err))
+			return
+		}
+
+		accountStatus, err := convert.MessageToAccountStatus(accountStatusResponse)
+		if err != nil {
+			sendErr(fmt.Errorf("error converting message to account status: %w", err))
+			return
+		}
+
```
peterargue, line 1259: before sending the response, I think this should also check that the message index is exactly one larger than the previous one received. if not return an error

access/grpc/grpc.go
```diff
@@ -1296,3 +1298,159 @@ func receiveBlocksFromClient[Client interface {
...
+		select {
+		case <-ctx.Done():
+			return
+		case accountStatutesChan <- accountStatus:
+		}
+	}
+}
+
+func checkAndIncrementMessageIndex(previousMessageIndex *atomic.Uint64, currentMessageIndex uint64) error {
```
illia-malachyn, line 1443: @peterargue is this implementation good enough? It might be good to add more tests for different msg indexes. What do you think?

access/grpc/grpc.go
```diff
@@ -1296,3 +1298,159 @@ func receiveBlocksFromClient[Client interface {
...
+	go func() {
+		defer close(accountStatutesChan)
+		defer close(errChan)
+		receiveAccountStatusesFromClient(ctx, subscribeClient, accountStatutesChan, errChan, &c.messageIndex)
+	}()
+
+	return accountStatutesChan, errChan, nil
+}
+
+func receiveAccountStatusesFromClient[Client interface {
```
peterargue, line 1396: using `Client` as the name is a little confusing, especially within a client library. how about breaking this out to a stand along type since it's used in a bunch of places, and renaming it to `Stream`?

access/grpc/grpc.go
```diff
@@ -1296,3 +1298,159 @@ func receiveBlocksFromClient[Client interface {
...
+	Recv() (*executiondata.SubscribeAccountStatusesResponse, error)
+}](
+	ctx context.Context,
+	client Client,
+	accountStatutesChan chan<- flow.AccountStatus,
+	errChan chan<- error,
+	previousMsgIndex *atomic.Uint64,
```
peterargue, line 1403: message index is specific to a stream. So when you start a new stream, the responses start with message index 0 (or 1, I don't remember how it started), then increment.

I think it's fine to track it only within this function as a regular uint64 value. There should not be any concurrent access.

access/grpc/grpc.go
peterargue: I think you can simply it a bit by tracking "next" instead of the current.
e.g.
```golang
next := 0 // since we start at 0, the first expected should be 0
for {
	if msg.MessageIndex != next {
		return err
	}
	next = msg.MessageIndex + 1
}
```

account.go
```diff
@@ -154,3 +154,19 @@ type accountKeyWrapper struct {
...
+
+type AccountStatus struct {
+	BlockID      Identifier
+	BlockHeight  uint64
+	MessageIndex uint64
+	Results      []*AccountStatusResult
+}
+
+type AccountStatusResult struct {
+	Address Identifier
```
peterargue, line 170: shouldn't this be an address?
```suggestion
	Address Address
```

peterargue:
added one small comment, but otherwise looks good

peterargue:
@illia-malachyn can you fix the conflicts, then this is ready to merge

peterargue (approved):

Issue #745, opened 2024-09-11
 Add SendAndSubscribeTransactionStatuses endpoint


Guitarheroua:
## Instructions 

Add `SendAndSubscribeTransactionStatuses` to the SDK that was implemented for Access API.

PR #772, opened 2024-09-25
Send and subscribe transaction statuses endpoint

AndriiDiachuk:
Closes: #745

### DIFF
--- a/access/grpc/grpc.go
+++ b/access/grpc/grpc.go
@@ -23,9 +23,11 @@ package grpc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
+	"github.com/onflow/flow/protobuf/go/flow/entities"
 	"google.golang.org/grpc"
 
 	"github.com/onflow/cadence"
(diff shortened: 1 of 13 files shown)


janezpodhostnik (approved):

access/grpc/grpc.go
```diff
@@ -1129,3 +1128,62 @@ func (c *BaseClient) subscribeEvents(
...
+			if err != nil {
+				if err == io.EOF {
+					// End of stream, return gracefully
+					return
+				}
+				sendErr(fmt.Errorf("error receiving blockHeader: %w", err))
+				return
+			}
+
+			txStatus := flow.TransactionStatus(txStatusResponse.GetTransactionResults().Status)
```
peterargue, line 1178: we should return the whole tx result, not just the status.

peterargue, line 1178: the response also includes a message index. this logic should track the index and make sure we get every message. if any message index is skipped, return an error indicating a response message was lost

peterargue (approved):

PR #773, opened 2024-09-26
Update GetNodeVersionInfo response object

### LINKED ISSUE #761: Update GetNodeVersionInfo response object
This is part of #736.

Align the response object we have in the protobuf schema with what we return in flow-go-sdk

The object we return on flow-go-sdk side:
https://github.com/onflow/flow-go-sdk/blob/master/access/grpc/grpc.go#L174

Protobuf schema:
https://github.com/onflow/flow/blob/master/protobuf/flow/entities/node_version_info.proto#L32 

Difference: 1 fields is missing.

```
// The compatible version range.
CompatibleRange compatible_range = 7;
```

illia-malachyn:
Closes #761

### DIFF
--- a/access/grpc/grpc.go
+++ b/access/grpc/grpc.go
@@ -166,10 +166,13 @@ func (c *BaseClient) GetNodeVersionInfo(ctx context.Context, opts ...grpc.CallOp
 		return nil, newRPCError(err)
 	}
 
+	var compRange *flow.CompatibleRange
 	info := res.GetInfo()
-	compRange := flow.CompatibleRange{
-		StartHeight: info.CompatibleRange.GetStartHeight(),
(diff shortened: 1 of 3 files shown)


janezpodhostnik (approved):

access/grpc/grpc.go
```diff
@@ -165,13 +165,19 @@ func (c *BaseClient) GetNodeVersionInfo(ctx context.Context, opts ...grpc.CallOp
 	}
 
 	info := res.GetInfo()
+	compRange := flow.CompatibleRange{
```
peterargue, line 168: it's possible for `CompatibleRange` to be nil if the version control isn't configured. let's make it a pointer here as well.

peterargue (approved):

Issue #768, opened 2024-09-24
CollectionGuarantee type differs from protobuf schema 


illia-malachyn:
This is a part of https://github.com/onflow/flow-go-sdk/issues/736

`CollectionGuarantee` is a part of `Collection` type which is used in `GetCollection` endpoint
https://github.com/onflow/flow-go-sdk/blob/master/access/grpc/grpc.go#L299

`CollectionGuarantee` type on the SDK side
https://github.com/onflow/flow-go-sdk/blob/master/collection.go#L47

`CollectionGuarantee` in the Protobuf schema
https://github.com/onflow/flow/blob/master/protobuf/flow/entities/collection.proto#L13

PR #779, opened 2024-10-15
align collection guarantee with protobuf schema

illia-malachyn:
Closes: #768


collection.go
```diff
@@ -45,7 +45,11 @@ func (c Collection) Encode() []byte {
 
 // A CollectionGuarantee is an attestation signed by the nodes that have guaranteed a collection.
 type CollectionGuarantee struct {
-	CollectionID Identifier
+	CollectionID     Identifier
+	Signatures       [][]byte
```
peterargue, line 49: you can skip this field. it's legacy from before we started using aggregate signatures, and is hardcoded to return the same value as `Signature`
https://github.com/onflow/flow-go/blob/ffaa335bacd8d5108ddf9a089278268e6ba10674/engine/common/rpc/convert/collections.go#L93

peterargue (approved):

Issue #763, opened 2024-09-20
Unify streaming endpoints code


illia-malachyn:
We need to do a small refactoring to get rid of code duplications caused by streaming endpoints introduced in #735 epic

Check out these PRs to understand the code similarities
https://github.com/onflow/flow-go-sdk/pull/762
https://github.com/onflow/flow-go-sdk/pull/760
https://github.com/onflow/flow-go-sdk/pull/758

illia-malachyn:
@peterargue cc

PR #781, opened 2024-10-17
refactor subscribe func and stream mock object

illia-malachyn:
Closes: #763

### DIFF
--- a/access/grpc/grpc.go
+++ b/access/grpc/grpc.go
@@ -1188,14 +1188,10 @@ func (c *BaseClient) SubscribeBlocksFromStartBlockID(
 		return nil, nil, newRPCError(err)
 	}
 
-	blocksChan := make(chan flow.Block)
-	errChan := make(chan error)
-
-	go func() {
-		defer close(blocksChan)
-		defer close(errChan)
-		receiveBlocksFromClient(ctx, subscribeClient, blocksChan, errChan)
-	}()
+	convertBlockResponse := func(response *access.SubscribeBlocksResponse) (flow.Block, error) {
+		return convert.MessageToBlock(response.GetBlock())
+	}
+	blocksChan, errChan := subscribe(ctx, subscribeClient.Recv, convertBlockResponse, "block")
 
 	return blocksChan, errChan, nil
 }
@@ -1221,14 +1217,10 @@ func (c *BaseClient) SubscribeBlocksFromStartHeight(
 		return nil, nil, newRPCError(err)
 	}
 
-	blocksChan := make(chan flow.Block)
-	errChan := make(chan error)
-
-	go func() {
-		defer close(blocksChan)
-		defer close(errChan)
-		receiveBlocksFromClient(ctx, subscribeClient, blocksChan, errChan)
-	}()
+	convertBlockResponse := func(response *access.SubscribeBlocksResponse) (flow.Block, error) {
+		return convert.MessageToBlock(response.GetBlock())
+	}
+	blocksChan, errChan := subscribe(ctx, subscribeClient.Recv, convertBlockResponse, "block")
 
 	return blocksChan, errChan, nil
 }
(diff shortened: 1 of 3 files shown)


illia-malachyn:
@peterargue cc

access/grpc/grpc.go
```diff
@@ -1744,52 +1567,58 @@ func (c *BaseClient) SubscribeBlockDigestsFromLatest(
...
-}](
+func subscribe[Response any, ClientResponse any](
 	ctx context.Context,
-	client Client,
-	blockDigestsChan chan<- flow.BlockDigest,
-	errChan chan<- error,
-) {
+	receive func() (*ClientResponse, error),
+	convertResponse func(*ClientResponse) (Response, error),
+	topicNameForErrors string,
```
peterargue, line 1582: you could just log the object's type instead of a string

access/grpc/grpc.go
```diff
@@ -1599,64 +1482,12 @@ func (c *BaseClient) SubscribeAccountStatusesFromLatestBlock(
...
-			return
-		}
-
-		accountStatus, err := convert.MessageToAccountStatus(accountStatusResponse)
-		if err != nil {
-			sendErr(fmt.Errorf("error converting message to account status: %w", err))
-			return
-		}
-
-		if accountStatus.MessageIndex != nextExpectedMsgIndex {
```
peterargue, line 1648: this check isn't included in `subscribe()`, but is still needed. you could probably generalize something that incorporates either messgeIndex or block height?

access/grpc/grpc.go
```diff
@@ -1744,52 +1567,58 @@ func (c *BaseClient) SubscribeBlockDigestsFromLatest(
...
+func subscribe[Response any, ClientResponse any](
 	ctx context.Context,
-	client Client,
-	blockDigestsChan chan<- flow.BlockDigest,
-	errChan chan<- error,
-) {
+	receive func() (*ClientResponse, error),
+	convertResponse func(*ClientResponse) (Response, error),
+	topicNameForErrors string,
+) (chan Response, chan error) {
```
peterargue, line 1583: only give the caller the receivers since this method is handling closing the channels
```suggestion
) (<-chan Response, <-chan error) {
```

access/grpc/grpc.go
```diff
@@ -1188,14 +1188,10 @@ func (c *BaseClient) SubscribeBlocksFromStartBlockID(
...
-
-	go func() {
-		defer close(blocksChan)
-		defer close(errChan)
-		receiveBlocksFromClient(ctx, subscribeClient, blocksChan, errChan)
-	}()
+	convertBlockResponse := func(response *access.SubscribeBlocksResponse) (flow.Block, error) {
+		return convert.MessageToBlock(response.GetBlock())
+	}
+	blocksChan, errChan := subscribe(ctx, subscribeClient.Recv, convertBlockResponse, "block")
```
peterargue, line 1194: you could reduce a little more boilerplate by having `subscribe` return a nil error and just returning the result directly
```suggestion
	return subscribe(ctx, subscribeClient.Recv, convertBlockResponse, "block")
```

peterargue, line 1194: or even something like this if you switch the error to use type
```suggestion
	return subscribe(ctx, subscribeClient.Recv, func(response *access.SubscribeBlocksResponse) (flow.Block, error) {
		return convert.MessageToBlock(response.GetBlock())
	})
```

peterargue (commented):
much cleaner. added a few comments, but generally this looks good

access/grpc/grpc.go
illia-malachyn: Hmm. I guess the most human-readable text would require sth like this.
`sendErr(fmt.Errorf("error receiving %s: %w", reflect.TypeOf(resp).Name(), err))`

Or do you have a better idea?

access/grpc/grpc.go
illia-malachyn: Actually, we return 3 values from the function `(chan, chan, error)`, so I cannot change it. 
I cannot even write `subscribe(), nil` due to language syntax restrictions

illia-malachyn: I actually like giving a name to a convert function as it looks more readable rather than an unnamed callback

peterargue: I mean change `subscribe` to return `(<-chan Response, <-chan error, error)`, where the error is always nil

access/grpc/grpc.go
peterargue: won't the type be something like `*access.SubscribeBlockDigestsResponse`, that's probably developer friendly enough. I'm also not attached to this.

access/grpc/grpc.go
illia-malachyn: This is a good catch. I came up with the second function for such cases. Please, take a look at it

peterargue (approved):

peterargue:
@illia-malachyn looks good. please merge master

PR #797, opened 2024-11-05
Use channel of pointers to objects for consistency

illia-malachyn:
I had to change the access interface of some endpoints. I'm not sure if I should've done this

### DIFF
--- a/access/client.go
+++ b/access/client.go
@@ -115,16 +115,16 @@ type Client interface {
 	GetExecutionDataByBlockID(ctx context.Context, blockID flow.Identifier) (*flow.ExecutionData, error)
 
 	// SubscribeExecutionDataByBlockID subscribes to execution data updates starting at the given block ID.
-	SubscribeExecutionDataByBlockID(ctx context.Context, startBlockID flow.Identifier) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error)
+	SubscribeExecutionDataByBlockID(ctx context.Context, startBlockID flow.Identifier) (<-chan *flow.ExecutionDataStreamResponse, <-chan error, error)
 
 	// SubscribeExecutionDataByBlockHeight subscribes to execution data updates starting at the given block height.
-	SubscribeExecutionDataByBlockHeight(ctx context.Context, startHeight uint64) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error)
+	SubscribeExecutionDataByBlockHeight(ctx context.Context, startHeight uint64) (<-chan *flow.ExecutionDataStreamResponse, <-chan error, error)
 
 	// SubscribeEventsByBlockID subscribes to events starting at the given block ID.
-	SubscribeEventsByBlockID(ctx context.Context, startBlockID flow.Identifier, filter flow.EventFilter, opts ...SubscribeOption) (<-chan flow.BlockEvents, <-chan error, error)
+	SubscribeEventsByBlockID(ctx context.Context, startBlockID flow.Identifier, filter flow.EventFilter, opts ...SubscribeOption) (<-chan *flow.BlockEvents, <-chan error, error)
 
 	// SubscribeEventsByBlockHeight subscribes to events starting at the given block height.
-	SubscribeEventsByBlockHeight(ctx context.Context, startHeight uint64, filter flow.EventFilter, opts ...SubscribeOption) (<-chan flow.BlockEvents, <-chan error, error)
+	SubscribeEventsByBlockHeight(ctx context.Context, startHeight uint64, filter flow.EventFilter, opts ...SubscribeOption) (<-chan *flow.BlockEvents, <-chan error, error)
 
 	// Close stops the client connection to the access node.
 	Close() error
--- a/access/grpc/client.go
+++ b/access/grpc/client.go
@@ -207,7 +207,7 @@ func (c *Client) GetTransactionResultsByBlockID(ctx context.Context, blockID flo
 func (c *Client) SendAndSubscribeTransactionStatuses(
 	ctx context.Context,
 	tx flow.Transaction,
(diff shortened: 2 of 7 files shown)


access/client.go
```diff
@@ -115,16 +115,16 @@ type Client interface {
 	GetExecutionDataByBlockID(ctx context.Context, blockID flow.Identifier) (*flow.ExecutionData, error)
 
 	// SubscribeExecutionDataByBlockID subscribes to execution data updates starting at the given block ID.
-	SubscribeExecutionDataByBlockID(ctx context.Context, startBlockID flow.Identifier) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error)
+	SubscribeExecutionDataByBlockID(ctx context.Context, startBlockID flow.Identifier) (<-chan *flow.ExecutionDataStreamResponse, <-chan error, error)
```
illia-malachyn, line 118: Should we change the interface too? Or I shouldn't touch these specific endpoints

access/grpc/convert/convert.go
```diff
@@ -193,7 +193,7 @@ func MessageToBlock(m *entities.Block) (flow.Block, error) {
 
 	tc, err := MessageToTimeoutCertificate(m.BlockHeader.GetLastViewTc())
 	if err != nil {
-		return flow.Block{}, err
+		return &flow.Block{}, err
```
peterargue, line 196: here and elsewhere 

```suggestion
		return nil, err
```

peterargue (commented):
added a comment about returning nil instead of an empty object. otherwise looks good

access/client.go
```diff
@@ -115,16 +115,16 @@ type Client interface {
...
 
 	// SubscribeExecutionDataByBlockHeight subscribes to execution data updates starting at the given block height.
-	SubscribeExecutionDataByBlockHeight(ctx context.Context, startHeight uint64) (<-chan flow.ExecutionDataStreamResponse, <-chan error, error)
+	SubscribeExecutionDataByBlockHeight(ctx context.Context, startHeight uint64) (<-chan *flow.ExecutionDataStreamResponse, <-chan error, error)
 
 	// SubscribeEventsByBlockID subscribes to events starting at the given block ID.
-	SubscribeEventsByBlockID(ctx context.Context, startBlockID flow.Identifier, filter flow.EventFilter, opts ...SubscribeOption) (<-chan flow.BlockEvents, <-chan error, error)
+	SubscribeEventsByBlockID(ctx context.Context, startBlockID flow.Identifier, filter flow.EventFilter, opts ...SubscribeOption) (<-chan *flow.BlockEvents, <-chan error, error)
```
illia-malachyn, line 124: @peterargue  We agreed that I can change `SubscribeExecutionDataByBlockID` interface.
Can I change this interface for events (`SubscribeEventsByBlockID`) too?

access/grpc/client.go
```diff
@@ -314,7 +314,7 @@ func (c *Client) SubscribeEventsByBlockID(
 	startBlockID flow.Identifier,
 	filter flow.EventFilter,
 	opts ...access.SubscribeOption,
-) (<-chan flow.BlockEvents, <-chan error, error) {
+) (<-chan *flow.BlockEvents, <-chan error, error) {
```
illia-malachyn, line 317: Same question here

access/client.go
peterargue: my main motivation with this request was to have the polling and subscription endpoints be consistent. With that said, let's leave the subscribe events as passed by value since the `GetEventsForHeightRange` endpoint also returns them by value.

peterargue (approved):

access/grpc/convert/convert.go
```diff
@@ -333,9 +333,9 @@ func BlockHeaderToMessage(b flow.BlockHeader) (*entities.BlockHeader, error) {
 	}, nil
 }
 
-func MessageToBlockHeader(m *entities.BlockHeader) (flow.BlockHeader, error) {
+func MessageToBlockHeader(m *entities.BlockHeader) (*flow.BlockHeader, error) {
 	if m == nil {
-		return flow.BlockHeader{}, ErrEmptyMessage
+		return &flow.BlockHeader{}, ErrEmptyMessage
```
janezpodhostnik, line 338: I think this can also be nil
```suggestion
		return nil, ErrEmptyMessage
```

access/grpc/convert/convert.go
```diff
@@ -346,10 +346,10 @@ func MessageToBlockHeader(m *entities.BlockHeader) (flow.BlockHeader, error) {
 
 	timeoutCertificate, err := MessageToTimeoutCertificate(m.GetLastViewTc())
 	if err != nil {
-		return flow.BlockHeader{}, fmt.Errorf("error converting timeout certificate: %w", err)
+		return &flow.BlockHeader{}, fmt.Errorf("error converting timeout certificate: %w", err)
```
janezpodhostnik, line 349: same here

access/grpc/convert/convert.go
```diff
@@ -423,12 +423,12 @@ func QuorumCertificateToMessage(qc flow.QuorumCertificate) (*entities.QuorumCert
 	}, nil
 }
 
-func MessageToBlockDigest(m *access.SubscribeBlockDigestsResponse) (flow.BlockDigest, error) {
+func MessageToBlockDigest(m *access.SubscribeBlockDigestsResponse) (*flow.BlockDigest, error) {
 	if m == nil {
-		return flow.BlockDigest{}, ErrEmptyMessage
+		return &flow.BlockDigest{}, ErrEmptyMessage
```
janezpodhostnik, line 428: and here

janezpodhostnik (commented):
Looks good, just some empty return types that could be nil instead.

janezpodhostnik (approved):

PR #798, opened 2024-11-27
[Access] Fix SendAndSubscribeTransactionStatuses and example

Guitarheroua:
Closes: [#6604](https://github.com/onflow/flow-go/issues/6604)

## Description

* Implements an example of `SendAndSubscribeTransactionStatuses` on the testnet, as the emulator does not support subscriptions.  
* The `messageIndex` value has been fixed, as it should always start at 1.

### DIFF
--- a/examples/send_and_subscribe_transaction_statuses/main.go
+++ b/examples/send_and_subscribe_transaction_statuses/main.go
@@ -21,6 +21,8 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/onflow/flow-go-sdk/crypto"
+	"log"
 
 	"github.com/onflow/flow-go-sdk/access/grpc"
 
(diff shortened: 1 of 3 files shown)


examples/send_and_subscribe_transaction_statuses/main.go
```diff
@@ -34,14 +35,33 @@ func main() {
...
 func demo() {
 	ctx := context.Background()
-	flowClient, err := grpc.NewClient(grpc.EmulatorHost)
+	flowClient, err := grpc.NewClient("access.testnet.nodes.onflow.org:9000")
 	examples.Handle(err)
 
-	serviceAcctAddr, serviceAcctKey, serviceSigner := examples.ServiceAccount(flowClient)
+	signerIndex := uint32(0)
+
+	signerPublicAddress := flow.HexToAddress("YOUR_ACCOUNT_ADDRESS")
```
chasefleming, line 43: Does this example work?

peterargue:
the issue here is that there are some AccessAPI endpoints that are missing from the emulator, which is why these changes were made. totally open to other options if you think there's a better way.

eventually we will update the emulator to add these endpoints, but there is no demand for it other than this test so far.

chasefleming (approved):

PR #868, opened 2025-07-04
Kan/passkeys compatibility

Kay-Zee:
Closes: #???

## Description

Adds initial understanding of the ExtensionData field for transaction signatures

Does not yet allow adding of ExtensionData to the signature, i.e. no passkeys support yet in this PR.

______

For contributor use:


transaction_test.go
```diff
@@ -126,7 +126,7 @@ func ExampleTransaction() {
 	// Envelope signatures:
 	// Address: ee82856bf20e2aa6, Key Index: 7, Signature: 03
 	//
-	// Transaction ID (after signing): d1a2c58aebfce1050a32edf3568ec3b69cb8637ae090b5f7444ca6b2a8de8f8b
+	// Transaction ID (after signing): d5d0f64f23650567657a7b2073f81df58e1eff3eb1c7af69eb83d65c4e25f4e5
```
tarakby, line 129: Shouldn't the transaction ID remain the same since this transaction has plain scheme signatures and our changes are backward compatible?

tarakby (commented):
- If my understanding is correct, the current transaction IDs in tests should remain the same after the change. Test should pass without updating the transaction IDs. 
- It would be good to add a new test where at least one transaction signature is of the webauthn scheme

transaction_test.go
tarakby: Actually the test passes with the original expected value, so we should just revert this change.

transaction_legacy.go
```diff
@@ -0,0 +1,141 @@
...
+ * limitations under the License.
+ */
+
+package flow
+
+import (
+	"bytes"
+	"errors"
+
+	"github.com/onflow/go-ethereum/rlp"
```
turbolent, line 25:
```suggestion
	"github.com/ethereum/go-ethereum/rlp"
```

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/onflow/cadence v1.7.0
 	github.com/onflow/crypto v0.25.3
 	github.com/onflow/flow/protobuf/go/flow v0.4.11
+	github.com/onflow/go-ethereum v1.16.2
```
turbolent, line 14: The fork should not be re-introduced, use upstream `"github.com/ethereum/go-ethereum/rlp"`

```suggestion
```

transaction.go
```diff
@@ -100,9 +101,13 @@ type payloadCanonicalForm struct {
 	Authorizers               [][]byte
 }
 
+type transactionSignatureCommonForm interface {
+	dummy()
```
turbolent, line 105: Use a descriptive name that is only implemented by intended types, e.g.

```suggestion
	isTransactionSignatureCommonForm()
```

transaction.go
```diff
@@ -569,31 +587,61 @@ type ProposalKey struct {
...
+func (s transactionSignatureCanonicalForm) dummy() {}
+
+// Checks if the scheme is plain authentication scheme, and indicate that it
+// is required to use the legacy canonical form.
+// We check for a valid scheme identifier, as this should be the only case
+// where the extension data can be left out of the cannonical form.
+// All other non-valid cases that are similar to the plain scheme, but is not valid,
+// should be included in the canonical form, as they are not valid signatures
+func (s TransactionSignature) shouldUseLegacyCanonicalForm() bool {
+	plainSchemeIdentifier := byte(0)
```
turbolent, line 615: This seems to be a constant for a magic number
```suggestion
	const plainSchemeIdentifier byte = 0
```

chasefleming:
@bjartek @bluesign Can you review as well? Will this break anything for you?

transaction.go
```diff
@@ -439,7 +450,18 @@ func (t *Transaction) Encode() []byte {
 func DecodeTransaction(transactionMessage []byte) (*Transaction, error) {
 	temp, err := decodeTransaction(transactionMessage)
 	if err != nil {
-		return nil, err
+		// If the transaction is in the legacy format, convert it to the canonical form
+		if strings.Contains(err.Error(), "too few elements") { // since the rlp library does not have this error type, just check string
```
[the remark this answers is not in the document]
Kay-Zee: not really, this is working off the go ethereum error: https://github.com/ethereum/go-ethereum/blob/master/rlp/decode.go#L108-L112

which is

```
type decodeError struct {
	msg string
	typ reflect.Type
	ctx []string
}
```

so only really gives you msg to work with. There is no additional ctx for the error we're working with.

chasefleming (approved):

bjartek:
Took a peek and cant see anything from our side.

bluesign:
all good from my side too, super excited for passkeys

